PR #307, opened 2026-03-21
Keeper is a state machine

ridafkih:


ridafkih:
Closing this, it was a largely-AI driven attempt to change Keeper.sh to use a more robust state machine as the underlying mechanism for handling the syncing engine. While I do think this is the right approach, without a much more well-defined set of "end goals," this has the potential to scope-creep forever.

I will tackle this more intelligently.

PR #383, opened 2026-05-18
fix(ics): don't wipe existing events when remote fetch fails

agurod42:
## Summary

When an ICS source's remote fetch fails (network blip, 5xx, TLS hiccup, DNS), `createIcsSourceFetcher` silently swallows the error in a `try/catch` and returns `{ events: [] }` — without the `unchanged: true` marker. `ingestSource` then takes that empty array as "the source authoritatively has zero events" and diffs against existing state, queuing **every existing `event_state` for deletion**. The next tick's flush wipes the calendar.

Verified in production for an Outlook-published ICS feed (313 VEVENT → 303 unique post-dedup): one successful fetch inserted 303 events; the very next tick failed at `pullRemoteCalendar`, the catch returned `events: []`, and the diff produced `events_removed: 303`. The destructive delete is logged as `outcome: success, events_added: 0, events_removed: 303` — no error visible to operators.

This is a silent data-loss bug for any ICS source behind a flaky endpoint.

## Why

`fetchRemoteIcal` was written to return `null` on error so the outer `fetchEvents` could short-circuit:

```ts
const fetchRemoteIcal = async (): Promise<string | null> => {
  try {
    const { ical } = await pullRemoteCalendar("ical", config.url, ...);
    return ical;
  } catch {
    return null;
  }
};

const fetchEvents = async (): Promise<FetchEventsResult> => {
  const ical = await fetchRemoteIcal();
  if (!ical) {
    return { events: [] };  // <- the bug
  }
  ...
};
```

`{ events: [] }` (no `unchanged: true`) is indistinguishable from "remote successfully returned an empty calendar". `ingestSource` then runs the normal diff path, which marks every row in `event_states` for removal.

For comparison, the OAuth source providers (`google`, `outlook`) propagate fetch errors and let the cron's outer `catch` log `outcome: error` and skip the flush. The ICS path was the odd one out.

## Changes

One source file edit + one new test file.

### `packages/calendar/src/ics/utils/fetch-adapter.ts`

Drops the `try/catch` so `pullRemoteCalendar` errors propagate. Adds a defensive empty-string check that returns `unchanged: true` instead of `events: []`, to preserve the no-wipe invariant if a future change ever lets an empty string slip through.

```ts
const fetchRemoteIcal = async (): Promise<string> => {
  const { ical } = await pullRemoteCalendar("ical", config.url, config.safeFetchOptions);
  return ical;
};

const fetchEvents = async (): Promise<FetchEventsResult> => {
  const ical = await fetchRemoteIcal();
  if (!ical) {
    // Defensive: pullRemoteCalendar already throws on invalid/empty bodies,
    // but if a future change ever returns an empty string here, treat it
    // as unchanged rather than authoritative-empty.
    return { events: [], unchanged: true };
  }
  ...
};
```

### `packages/calendar/tests/ics/utils/fetch-adapter.test.ts` (new)

Three cases, mocking `pull-remote-calendar` and `create-snapshot` so the adapter is exercised end-to-end without network or DB:

1. **Regression** — when `pullRemoteCalendar` rejects, `fetchEvents` propagates the error and `createSnapshot` is never called. This is the case that previously caused the wipe.
2. Happy path — a successful changed fetch returns parsed events.
3. Unchanged path — `createSnapshot` returning `changed: false` produces `{ events: [], unchanged: true }`.

## Testing

`bun test` clean across `packages/calendar` (336/336 passing, including the 3 new cases).

`bun run types` clean across `packages/calendar`, `services/api`, `services/cron`, `applications/web` on this change.

## Compatibility

- No schema changes.
- No public API changes.
- Behaviour change is strictly safer: on a transient fetch failure, the cron now logs `outcome: error` and skips the flush entirely, leaving existing `event_state` rows intact for the next retry. This matches what the OAuth providers already do.

## Out of scope

- **Retry-on-error in the cron.** The cron already classifies the error as `retriable: true` but doesn't currently re-attempt within a tick. Adding a bounded retry would further reduce noise but is independent of the data-loss fix.
- **`failureCount` bookkeeping on the source.** Touched by other paths; not in this PR.


ridafkih (approved):
This looks great, as long as actual empty calendars are properly propagated upstream, we're good! Seems to be the case since it's checking the actual string value for truthiness.

PR #385, opened 2026-05-19
fix(ics): strip stale VTIMEZONE blocks for IANA zones

agurod42:
## Summary

Some CalDAV servers — notably iCloud — embed `VTIMEZONE` blocks whose recurrence rules are truncated at the year DST was last applied. For zones that abolished DST after that cutoff, the embedded rules extrapolate DST forward forever, yielding the wrong UTC offset for any date after the cutoff.

Concrete case: Uruguay (`America/Montevideo`) eliminated DST in 2015. iCloud's `VTIMEZONE` for that zone has its last `BEGIN:DAYLIGHT` ending `RRULE...UNTIL=20141005` and last `BEGIN:STANDARD` ending `UNTIL=20150308`. After that, the block defines no rules. `ts-ics` keeps applying the last DAYLIGHT transition (Oct → Mar) forward indefinitely, so an event at `2026-05-19T08:00:00 America/Montevideo` parses as `2026-05-19T10:00:00Z` (UTC-2, applying phantom DST) instead of the correct `2026-05-19T11:00:00Z` (UTC-3).

The same pattern affects any zone that abolished DST after the year iCloud generated its VTIMEZONE truncation — and it's not unique to iCloud; any server that ships a historical VTIMEZONE without future rules has this risk.

## Fix

When the `TZID` matches a runtime-recognized IANA zone (`new Intl.DateTimeFormat({ timeZone: tzid })` doesn't throw), strip the `VTIMEZONE` block so `ts-ics` falls back to the platform's IANA tzdata (Node/Bun's ICU), which has accurate post-abolition rules.

Non-IANA TZIDs are left untouched — the embedded `VTIMEZONE` is the only way to resolve them, so we can't strip it without breaking those events:
- Microsoft Windows zone names like `"Eastern Standard Time"`, `"Pacific Standard Time"`.
- Custom/private labels that some calendar servers emit.
- Anything `Intl.DateTimeFormat` rejects.

The strip happens at the `parseIcsCalendar` wrapper, so both CalDAV and ICS providers benefit.

## Repro

The added regression test exercises iCloud's exact serialization for `America/Montevideo`. Without the fix, `event.start.date.toISOString()` returns `"2026-05-19T10:00:00.000Z"`; with it, the correct `"2026-05-19T11:00:00.000Z"`.

## Notes

- One unit test in `parse-ics-calendar.test.ts` that asserted "throws on malformed VTIMEZONE" had to change its TZID from `America/New_York` (now stripped) to a non-IANA TZID (still parsed). The behavior under test — that malformed VTIMEZONE breaks parsing for non-IANA zones — is preserved.
- No schema, dependency, or runtime changes beyond the strip helper.

### DIFF
--- a/packages/calendar/src/ics/utils/parse-ics-calendar.ts
+++ b/packages/calendar/src/ics/utils/parse-ics-calendar.ts
@@ -4,7 +4,46 @@ interface ParseIcsCalendarOptions {
   icsString: string;
 }
 
-const parseIcsCalendar = (options: ParseIcsCalendarOptions) =>
-  convertIcsCalendar(globalThis.undefined, options.icsString);
+const VTIMEZONE_BLOCK_PATTERN = /BEGIN:VTIMEZONE\r?\n([\s\S]*?)END:VTIMEZONE\r?\n?/g;
+const TZID_LINE_PATTERN = /(?:^|\r?\n)TZID:(.+?)(?:\r?\n|$)/;
 
-export { parseIcsCalendar };
+const isIanaTimezone = (tzid: string): boolean => {
+  try {
+    // Throws RangeError for non-IANA values (including Windows zone names like
+    // "Eastern Standard Time"). The runtime falls back to a static lookup, so
+    // this is O(1) — no need to memoize.
+    new Intl.DateTimeFormat("en", { timeZone: tzid });
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+/**
+ * iCloud (and some other CalDAV servers) embed a VTIMEZONE block that lists
+ * historical DST transitions but truncates future rules at the year the zone
+ * was last updated. For zones like America/Montevideo — which abolished DST in
+ * 2015 — the embedded VTIMEZONE has no rules after 2015, and ts-ics extrapolates
+ * the last DAYLIGHT rule forward indefinitely, producing UTC offsets that are
+ * off by one hour for any date after the last rule.
+ *
+ * When the TZID matches a runtime-recognized IANA zone, we strip the VTIMEZONE
+ * block so ts-ics falls back to the platform's IANA tzdata (Node/Bun's ICU),
+ * which has correct post-2015 rules. Non-IANA TZIDs (Microsoft Windows zone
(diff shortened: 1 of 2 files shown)


ridafkih:
@agurod42 When you worked on this, did you consider whether it'd be worth it to try to bound "trusting" the server? We're now unconditionally overriding the ICS-provided information, which may be okay - just wondering if it's something that was given thought. This does fix it for iCloud, but the fun thing about working with these is every provider (especially Outlook/Microsoft) loes to do their own funky little implementation and spec-violation. :P 

Just want to make sure we don't end up landing a regression.

packages/calendar/src/ics/utils/parse-ics-calendar.ts
```diff
@@ -4,7 +4,46 @@ interface ParseIcsCalendarOptions {
   icsString: string;
 }
 
+const VTIMEZONE_BLOCK_PATTERN = /BEGIN:VTIMEZONE\r?\n([\s\S]*?)END:VTIMEZONE\r?\n?/g;
```
ridafkih, line 7: Is there a way to handle this using ts-ics? I'd rather avoid Regex parsing as much as possible.

packages/calendar/src/ics/utils/parse-ics-calendar.ts
```diff
@@ -4,7 +4,46 @@ interface ParseIcsCalendarOptions {
...
+const TZID_LINE_PATTERN = /(?:^|\r?\n)TZID:(.+?)(?:\r?\n|$)/;
+
+const isIanaTimezone = (tzid: string): boolean => {
+  try {
+    // Throws RangeError for non-IANA values (including Windows zone names like
+    // "Eastern Standard Time"). The runtime falls back to a static lookup, so
+    // this is O(1) — no need to memoize.
+    new Intl.DateTimeFormat("en", { timeZone: tzid });
```
ridafkih, line 15: I know this is a notoriously slow surface area, wondering if this will have any performance hits with large ICS. If we're just testing for Microsoft's funkiness, we could just check against that list - might then be worth turning it into a Set.

PR #386, opened 2026-05-20
fix(ics): emit RRULE and EXDATE for recurring events in user feed

agurod42:
## Summary

The user-facing ICS feed at `/api/cal/{token}.ics` emits recurring events as single one-off `VEVENT`s anchored at the master `DTSTART`. Two gaps converge:

1. `generateUserCalendar` (in `services/api/src/utils/ical.ts`) does not select `recurrenceRule` or `exceptionDates` from `event_states`, so the data never reaches the formatter.
2. `formatEventsAsIcal` (in `services/api/src/utils/ical-format.ts`) does not thread `recurrenceRule` or `exceptionDates` into the `IcsEvent` it passes to `ts-ics`, so even if they arrive they get dropped.

Net effect: a weekly meeting set up in October 2025 appears in client calendars as a single event in October 2025 and never again — every recurring event becomes a single occurrence in the distant past.

## Fix

- Adds `recurrenceRule` and `exceptionDates` to the `event_states` SELECT in `generateUserCalendar`.
- Adds a small helper that `JSON.parse`s the stored JSON strings (the providers store these via `JSON.stringify(...)`) and revives any ISO-string dates back into `Date` objects — `ts-ics` expects `Date` instances on the in-memory shape (`recurrenceRule.until.date`, each `exceptionDates[i].date`).
- Adds `recurrenceRule` and `exceptionDates` to the `CalendarEvent` interface in `ical-format.ts`, and sets them on the emitted `IcsEvent` when present.

## Repro / tested

Against a CalDAV calendar with an `RRULE:FREQ=WEEKLY` master set up months ago, the feed previously had zero `RRULE:` lines despite many recurring events being stored. After the fix the same calendar produces `RRULE:` for each master plus `EXDATE` for events with skipped occurrences.

Added unit tests in `services/api/tests/utils/ical.test.ts`:
- `emits RRULE for events with a recurrenceRule` — covers the basic case.
- `emits EXDATE for events with exceptionDates` — covers skipped occurrences.
- `does not emit RRULE for one-off events` — guards against accidentally emitting RRULE/EXDATE when neither is set.

## Notes

- No schema, migration, or runtime changes; purely additive.
- The provider parsers (Google/Zoho/Outlook/CalDAV/ICS) already capture `recurrenceRule` and `exceptionDates` and persist them to `event_states` via `serializeOptionalJson` / `stringifyIfPresent`. This change wires the read path so the feed actually emits them.


ridafkih (approved):

services/api/src/utils/ical.ts
```diff
@@ -1,9 +1,56 @@
...
 import { formatEventsAsIcal } from "./ical-format";
-import type { FeedSettings } from "./ical-format";
+import type { CalendarEvent, FeedSettings } from "./ical-format";
+
+const parseJsonField = <T>(value: string | null): T | null => {
+  if (!value) {
+    return null;
+  }
+  try {
+    return JSON.parse(value) as T;
```
ridafkih: We should really be validating this, rather than using type assertions. You can use a predicate to verify the structure.

services/api/src/utils/ical.ts
```diff
@@ -1,9 +1,56 @@
...
+};
+
+const parseRecurrenceRule = (value: string | null): IcsRecurrenceRule | null => {
+  const parsed = parseJsonField<IcsRecurrenceRule>(value);
+  if (!parsed) {
+    return null;
+  }
+  // ts-ics expects Date objects on the in-memory shape (e.g. UNTIL is a {date: Date}).
+  // JSON.parse leaves these as ISO strings, so revive them before re-serializing.
+  return reviveDates(parsed) as IcsRecurrenceRule;
```
ridafkih, line 43: Ditto.

services/api/src/utils/ical.ts
```diff
@@ -1,9 +1,56 @@
...
+};
+
+const parseExceptionDates = (value: string | null): IcsDateObject[] | null => {
+  const parsed = parseJsonField<IcsDateObject[]>(value);
+  if (!parsed) {
+    return null;
+  }
+  return reviveDates(parsed) as IcsDateObject[];
```
ridafkih, line 51: Ditto;.

services/api/src/utils/ical.ts
```diff
@@ -1,9 +1,56 @@
...
+  }
+  try {
+    return JSON.parse(value) as T;
+  } catch {
+    return null;
+  }
+};
+
+const reviveDates = (value: unknown): unknown => {
+  if (typeof value === "string" && /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?Z$/.test(value)) {
```
ridafkih, line 21: Would prefer if the Regex pattern here was pulled up to the top, but I tend to try to avoid patterns like this as much as possible.

ridafkih (changes requested):
Left a few comments above, may address them myself.

ridafkih (approved):

PR #387, opened 2026-05-20
fix(ics): emit recurring events as master + RECURRENCE-ID overrides

agurod42:
## Summary

When a CalDAV/ICS source serializes a recurring event with modified instances (e.g. an Outlook weekly meeting where one occurrence has been moved to a different time), each override arrives as a `VEVENT` with the **same UID as the master** plus a `RECURRENCE-ID` pointing at the occurrence being replaced.

Keeper parses each override into its own `event_states` row but **drops the `RECURRENCE-ID`**, then emits them back into the user feed as standalone `VEVENT`s with brand-new UIDs. Calendar clients then show **both** the original occurrence (from the master's RRULE expansion) **and** the override as a separate event — a visible duplicate.

> Stacked on top of #386 (RRULE/EXDATE emission). Merge #386 first; the changes here build directly on that base. The first commit on this branch is the same as #386, so the only "new" commit reviewed here is `b6435f67`.

## Reproduction

1. Subscribe to an Outlook .ics feed that contains a weekly meeting with a single overridden occurrence (e.g. user moves one Wednesday's meeting from 10:00 to 14:30).
2. Subscribe a calendar client (macOS Calendar / Apple Calendar / Google) to the Keeper feed and inspect the corresponding week.

**Before this fix:** the week shows two events — the master at 10:00 (from RRULE expansion) AND a separate one-off at 14:30 (the override emitted with a new UID).

**After this fix:** the client sees a single occurrence at 14:30, because the override carries the master's UID + a `RECURRENCE-ID` matching the original 10:00 slot, so the client correctly hides the original.

### 1. New `recurrenceId` column

`packages/database/drizzle/0070_event_states_recurrence_id.sql`:

```sql
ALTER TABLE "event_states" ADD COLUMN "recurrenceId" timestamp;
```

Existing rows have `NULL`; subsequent ingests populate the column for overrides.

### 2. Parsers read from `ts-ics`'s shape

The `ts-ics` `IcsRecurrenceId` type is:

```ts
type IcsRecurrenceId = { range?: "THISANDFUTURE"; value: IcsDateObject };
```

So the parsed event exposes the date at `event.recurrenceId.value.date`, NOT `event.recurrenceId.date`. Both the ICS parser (`packages/calendar/src/ics/utils/parse-ics-events.ts`) and the CalDAV parser (`packages/calendar/src/providers/caldav/shared/ics.ts`) now read from `.value.date`.

### 3. Threaded through the ingest pipeline

Added `recurrenceId?: Date` to `SourceEvent`, `SyncableEvent`, `EventTimeSlot`, and `ParsedCalendarEvent`. The shared ingest flush in `services/cron/src/jobs/ingest-sources.ts` and the CalDAV provider's DB write include `recurrenceId` in the insert. `EVENT_STATE_CONFLICT_SET` in `packages/calendar/src/core/source/write-event-states.ts` includes the column so re-syncs update it.

### 4. Feed generator groups + emits `RECURRENCE-ID`

`services/api/src/utils/ical.ts` selects `recurrenceId` and `sourceEventUid`. `services/api/src/utils/ical-format.ts` groups rows by `sourceEventUid` and emits:

- **The master** (the row with `recurrenceRule != null` and `recurrenceId == null`) with its own derived UID and the `RRULE`/`EXDATE`.
- **Each override** under the same UID as the master, with `RECURRENCE-ID` set to its original-occurrence timestamp.

The `RRULE` and `EXDATE` are only emitted on the master itself, never on overrides.

### 5. Regression tests

`services/api/tests/utils/ical.test.ts` adds:

- `groups master + overrides under a single UID with RECURRENCE-ID`
- `emits standalone UIDs for events lacking sourceEventUid`

These pin the new behavior so future refactors can't silently regress it.

## Verified

Tested end-to-end against a real Outlook .ics feed containing several modified instances of a weekly meeting. Before: clients saw the weekly slot plus 16 stray standalone events at modified times. After: clients see exactly the right occurrences — masters expand cleanly, modified instances appear in their new slots and replace (rather than supplement) the original.

## Notes

- New rows get `recurrenceId` populated on next ingest. For existing rows from before the migration, force a re-sync (e.g. delete the relevant `event_states` rows or clear the snapshot) to populate the field. We won't backfill automatically since the source `RECURRENCE-ID` isn't stored anywhere recoverable.
- No breaking changes to the API surface; new column is nullable and additive everywhere.


ridafkih:
@agurod42 Well done, I'll be reviewing your contributions in the next 24 hours! Thanks so much! I've already skimmed and they look great, am looking forward to taking a deeper dive. This one in particular is such a big help!

services/api/src/utils/ical-format.ts
```diff
@@ -43,6 +47,70 @@ const resolveEventSummary = (event: CalendarEvent, settings: FeedSettings): stri
...
+    }
+    const masterIdx = group.findIndex((e) => e.recurrenceRule !== null && e.recurrenceId === null);
+    if (masterIdx > 0) {
+      const [master] = group.splice(masterIdx, 1);
+      group.unshift(master);
+    }
+    result.push(group);
+  }
+  return [...result, ...singletons];
+};
```
ridafkih: This is doing a lot of iterations, wonder if we could trim this down? At the very least, it'd be nice to split it up into well-named, granular functions.

ridafkih (changes requested):
This is really great, left a few nits.

ridafkih (approved):

PR #403, opened 2026-06-02
fix(google): write events via events.import instead of insert

ridafkih:


packages/calendar/src/providers/google/destination/provider.ts
```diff
@@ -96,6 +73,7 @@ const createGoogleSyncProvider = (config: GoogleSyncProviderConfig) => {
 
   const eventsPath = `/calendar/v3/calendars/${encodeURIComponent(config.externalCalendarId)}/events`;
 
+  // Writes go through events.import, which upserts by iCalUID: re-pushing an existing event updates it rather than 409ing.
```
ridafkih:
```suggestion
```

packages/calendar/src/providers/google/destination/provider.ts
```diff
@@ -402,7 +232,8 @@ const createGoogleSyncProvider = (config: GoogleSyncProviderConfig) => {
 
       if (deleteResponse.statusCode >= 200 && deleteResponse.statusCode < 300) {
         results[originalIndex] = { success: true };
-      } else if (deleteResponse.statusCode === HTTP_STATUS.NOT_FOUND) {
+      } else if (deleteResponse.statusCode === HTTP_STATUS.NOT_FOUND || deleteResponse.statusCode === GONE_STATUS) {
+        // 404 (never existed) and 410 (already deleted) both mean the event is gone — the desired end state.
```
ridafkih:
```suggestion
```

Issue #137, opened 2026-02-24
Bug: Updates/deletes on source calendars don’t sync to destination (self-hosted, Google+Microsoft via OAuth)


ckoklas:
Hi Rida. First of all, thank you for building Keeper! 🙏

I’m self-hosting Keeper (keeper-standalone) and syncing from multiple source calendars (Google + Microsoft, all connected via OAuth) into a Microsoft destination calendar. New events get created in the destination calendar, but changes and deletions on the source calendars do not get reflected in the destination calendar.

**What I’m seeing**
- If an event is deleted on a source calendar, it remains in the destination calendar.
- If an event’s time is changed on a source calendar, the destination calendar does not update (it stays at the original time from when it was first added).

ridafkih:
@ckoklas Interesting, will have a look at this today. Thanks so much for filing the report! I'll let you know if I need any more information from you or if I have trouble reproducing the issue.

ridafkih:
@ckoklas Can you check if the syncing now works on the hosted version?
This appears to be a transient issue, with silent failures over time. Restarting the managed instance in the cloud resolved it - investigating the root cause later today.

ridafkih:
I think I've narrowed it down to https://github.com/oven-sh/bun/issues/24019, considering switching to `ioredis` in its place.

ridafkih:
Yeah, this definitely has to do with cron locks. Provided is a graph, and you can see that free & pro syncs halt completely indiscriminately. The darker blue one is pro syncing. Sometimes it happens quickly, other times - after a day or two. There are two possibilities. 

Something is causing Redis to hang, and my suspicion is that it's Bun's RedisClient. I'm patching in `ioredis` and will observe. 

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/21001b64-f64b-4f00-b418-c2f93d520fc9" />

mbaroody:
this is still happening for me on 2.9.13. however, it seems to only be happening when a recurring event is declined/rescheduled on a one-time basis. but i have the opposite situation: two Microsoft ics calendars set to sync to a Google Calendar

PR #409, opened 2026-06-11
fix(outlook): hydrate recurring occurrences from series master during ingestion

### LINKED ISSUE #407: M365: Recurring entries only get the first entry synced
When syncing M365 calendars, I have lots of recurring entries for every day and week. Using keeper, only the first entry of every chain is synced, the rest is ignored.

I already implemented a fix on my local instance and currently testing it out and debugging it for edge cases I overlooked. 

If it's alright with you, I would propose the fix via a PR as soon as I'm confident it's actually a robust, complete and slim fix

valentindiehl:
Fixes #407

## Problem

Recurring M365 events only ever synced as a single event. Graph's `calendarView/delta` expands recurring series server-side, but each expanded `occurrence` item carries only `id`, `type`, `seriesMasterId`, `start`, and `end` — no `iCalUId`, subject, or anything else. `parseOutlookEvents` dropped them for the missing `iCalUId`, so the only row that survived was the series master itself, which Graph lists with the *first* occurrence's start/end. (This is also where the ancient ghost events come from — e.g. a yearly series created in 2012 shows up once, on its 2012 date.)

## Fix

- Request `type` and `seriesMasterId` in `$select` and allow them in the event schema (`string | null`, Graph sends explicit nulls).
- After fetching all pages, hydrate each bare occurrence from its series master: metadata from the master, times from the occurrence.
- Drop the master's own row — its start/end just duplicates the first occurrence.
- If an occurrence's master isn't in the payload (delta syncs, or series that started before the sync window), fetch it via `GET /me/events/{id}`. A 404 means the series was deleted in the meantime; the orphaned occurrence is skipped.
- Bump `OUTLOOK_SYNC_TOKEN_VERSION` so existing calendars do a full re-sync and backfill occurrences that were previously dropped.

The hydration gate is `seriesMasterId && !iCalUId` rather than `type === "occurrence"`, so `exception` items (modified occurrences) pass through untouched when Graph delivers them with full properties, and still get hydrated with their own (correct) times if a stub ever shows up.

Storage-wise this is safe: all occurrences of a series share the master's `iCalUId`, but event states are keyed on `(calendarId, sourceEventUid, startTime, endTime)`, so they coexist and series edits update in place.

## Known limitation

Delta `@removed` entries carry only the Graph `id`, while stored rows are keyed by `iCalUId` — so cancelling a single occurrence mid-series doesn't remove its row until the next full re-sync (delta token expiry / 410). That mismatch predates this PR and affects non-recurring events too (looks like #137); fixing it properly needs the Graph id persisted alongside the uid, which felt like scope creep here.

## Verification

- Ran against a real M365 tenant: a daily series that previously synced as one event produced 75 occurrences within the sync window; 197 blockers pushed to destination calendars with 0 failures. The 2012-dated ghosts disappeared after the version-bump re-sync.
- New tests cover: hydration from an in-payload master, fetching a master missing from a delta payload, skipping occurrences whose master 404s, and exceptions with full properties passing through unchanged.
- `bun run types`, `bun run lint`, `bun run unused`, and `bun run test` all pass.


ridafkih:
Closing in favour of #443. Your changes _were_ used as the foundation for that change.

PR #411, opened 2026-06-13
fix(caldav): convert HTML descriptions to plain text and stash original under X-ALT-DESC

lucky-lusa:
## Summary

iCloud (and other strict CalDAV servers) rejects PUTs whose `DESCRIPTION` property carries raw HTML, returning `404 Not Found` for Outlook-style Teams meeting bodies and similar imports. The destination provider was also reading the response body without checking `response.ok`, so those rejections looked like successes and stale mappings persisted.

## Changes

- `eventToICalString`: detect HTML in the source description, convert it to plain text while preserving anchor URLs inline as `Text (URL)`, and emit the original HTML as `X-ALT-DESC;FMTTYPE=text/html` with RFC 5545 line folding (≤75 octets). Plain-text descriptions still pass through unchanged.
- `CalDAVClient.createCalendarObject`: throw with the response status, statusText and URL when iCloud answers with a non-2xx so the sync layer records a real failure instead of optimistically storing a mapping.

## Test plan
- [ ] Push an event with an Outlook Teams meeting description (HTML body, long meetup-join URL): iCloud returns 201 and the event becomes visible in the destination calendar.
- [ ] Push an event with a plain-text description: ICS output is unchanged (no `X-ALT-DESC`).
- [ ] Push an event the server actively rejects: error is propagated up rather than treated as success.
- [ ] Re-sync after the first run: existing mappings remain, no duplicate PUT.

### DIFF
--- a/packages/calendar/src/providers/caldav/shared/ics.ts
+++ b/packages/calendar/src/providers/caldav/shared/ics.ts
@@ -46,11 +46,67 @@ const resolveEventTimezones = (
   return [timezone];
 };
 
+const HTML_TAG_PATTERN = /<[^>]*>/g;
+const HTML_ENTITIES: Record<string, string> = {
+  "&nbsp;": " ",
+  "&amp;": "&",
+  "&lt;": "<",
+  "&gt;": ">",
+  "&quot;": "\"",
+  "&apos;": "'",
+  "&#39;": "'",
+};
+const ANCHOR_PATTERN = /<a\b[^>]*?href="([^"]*)"[^>]*?>([\s\S]*?)<\/a>/gi;
+const ICS_LINE_MAX_OCTETS = 75;
+
+const isHtmlDescription = (text: string): boolean => /<\/?\w+[^>]*>/.test(text);
+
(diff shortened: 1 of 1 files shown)


ridafkih:
@lucky-lusa I'm leaning towards ensuring that we're using plaintext descriptions, if available, from the sources rather than parsing and cleansing HTML in rich-text descriptions we get back. If we were to do the latter, however, I think it'd be much cleaner to `new JSDOM(description).document.body.textContent` rather than trying to parse it so surgically. :) 

investigating & making the adjustments

lucky-lusa:
Closing — superseded by #750 (`toPlainTextDescription` / `normalizeCalDAVEvent`), which lands upstream's own, more complete DOM-aware HTML-to-plain-text conversion applied once at the normalize layer. This PR's regex-based approach inside `eventToICalString` (plus the X-ALT-DESC stash) is now redundant and, if merged as-is, would double-apply the projection on top of #750's normalization. We've since dropped this approach on our side in favor of #750.

PR #452, opened 2026-07-20
feat: configurable historic and future sync windows

ridafkih:
Adds a per-calendar historic and future sync range (1 week, 1 month, 3, 6, 12 months, 2 years), defaulting to one month back and two years forward. Each source is ingested across the widest window its mapped destinations need and records the coverage it achieved, so a widened window never becomes authoritative before the data behind it exists. Mirrors falling outside the window are retired at both edges, while ICS sources keep unbounded stored history — the sync window governs what propagates to destination calendars, not what Keeper retains and shows you.

Adversarial review found and fixed a number of defects in the original implementation:

- adds were matched only against authoritative mappings, so a mapping between recorded coverage and the requested edge looked unmapped and churned one create plus one delete per event, per run, indefinitely
- the plan clamp was re-applied on every sync, so any non-active subscription status (a card retry, a trial) shrank a paying user's window and deleted the history outside it; only Pro can store a non-default range, so the clamp is gone
- an over-budget recurrence series failed the whole calendar's ingestion, then the destination read, then leaked its detached occurrences; it is now withheld and its existing mirrors are spared
- migrations aborted on databases at 0070-0074 by naming a column added in 0075, and the source-calendar backfill rescanned rows it had already filled
- unbounded ICS history removed the only thing bounding a single statement, so both the event-state insert and delete now chunk under the bind-parameter ceiling
- destination pushes could livelock: every enqueue now dedups on a stable job id, and a lock handle no longer stays poisoned after one transient renewal failure

Verified the upgrade from every released schema state on live Postgres, twice each, plus a live end-to-end sync from an ICS feed into a CalDAV destination. On production data the first sync retires 144 aged-out mirrors and the migration runs in about two seconds.


applications/web/src/routes/(dashboard)/dashboard/accounts/$accountId.$calendarId.tsx
```diff
@@ -158,13 +167,124 @@ function CalendarDetailPage() {
...
+  locked,
+}: {
+  calendarId: string;
+  field: SyncRangeField;
+  label: string;
+  locked: boolean;
+}) {
+  const store = useStore();
+  const calendar = useAtomValue(calendarDetailAtom);
+  const fallbackRange: SyncRange = field === "syncHistoricRange" ? "1_week" : "2_years";
```
ridafkih: Why is a `fallbackRange` even needed here, and why does it depend on the field value?

applications/web/src/routes/(dashboard)/dashboard/accounts/$accountId.$calendarId.tsx
```diff
@@ -158,13 +167,124 @@ function CalendarDetailPage() {
...
+  const selectedRange = calendar?.[field] ?? fallbackRange;
+
+  const selectRange = (range: SyncRange) => {
+    if (locked || range === selectedRange) return;
+
+    track(ANALYTICS_EVENTS.calendar_setting_toggled, { field, value: range });
+    store.set(calendarDetailAtom, (previous) => (
+      previous ? { ...previous, [field]: range } : previous
```
ridafkih, line 239: Just use a guard clause.

packages/calendar/src/core/sync/operations.ts
```diff
@@ -451,20 +480,22 @@ const computeSyncOperations = (
...
-  const mappingIdsToPrune = existingMappings.flatMap((mapping) => {
-    if (localEventIds.has(getMappingSyncEventId(mapping))) {
-      return [];
-    }
-    const remoteExists = remoteEventsByMappingId.has(mapping.id);
-    if (mapping.endTime < timeBoundary.syncWindowStart && !remoteExists) {
-      return [mapping.id];
+  const mappingIdsToPrune: string[] = [];
+  const syncWindowEnd = timeBoundary.syncWindowEnd
+    ?? new Date(8_640_000_000_000_000);
```
ridafkih: Do we really expect a fallback to be used here?

packages/calendar/src/core/sync-engine/ingest.ts
```diff
@@ -102,7 +111,8 @@ const ingestSource = async (options: IngestSourceOptions): Promise<IngestionResu
     const withPersistenceTransaction = resolvePersistenceTransaction(options);
     const fetchResult = await fetchEvents();
     wideEvent["source_events.count"] = fetchResult.events.length;
-    const recurrenceValidationWindow = getOAuthSyncWindow(RECURRENCE_VALIDATION_YEARS);
+    const recurrenceValidationWindow = fetchResult.syncWindow
```
ridafkih: This is meant to resolve recurring master events that fall outside of our window, no? Why would we use the `syncWindow` here for that?

packages/calendar/src/core/sync-engine/ingest.ts
```diff
@@ -166,6 +176,18 @@ const ingestSource = async (options: IngestSourceOptions): Promise<IngestionResu
...
+            return [];
+          }
+          if (
+            event.endTime <= fetchResult.syncWindow.timeMin
+            || event.startTime >= fetchResult.syncWindow.timeMax
+          ) {
+            return [event.id];
+          }
+          return [];
+        }),
```
ridafkih: Rather than having together a sloppy flatMap for this, just use a function with a for-loop inside?

packages/calendar/src/core/sync/operations.ts
```diff
@@ -397,21 +413,26 @@ const buildRemoveOperations = (
...
 ): SyncOperation[] => {
   const operations: SyncOperation[] = [];
-  const remoteIdentities = new Set(
-    remoteEvents.map((remoteEvent) => `${remoteEvent.uid}\u0000${remoteEvent.deleteId}`),
-  );
+  const syncWindowEnd = timeBoundary.syncWindowEnd ?? new Date(8_640_000_000_000_000);
+  const cleanupWindowStart = timeBoundary.cleanupWindowStart
+    ?? new Date(-8_640_000_000_000_000);
+  const cleanupWindowEnd = timeBoundary.cleanupWindowEnd
+    ?? new Date(8_640_000_000_000_000);
```
ridafkih: A mess.

packages/calendar/src/core/sync/sync-range.ts
```diff
@@ -0,0 +1,92 @@
...
+    const shifted = new Date(date);
+    shifted.setDate(shifted.getDate() + direction * 7);
+    return shifted;
+  }
+
+  const months = SYNC_RANGE_MONTHS[range];
+  if (typeof months !== "number") {
+    throw new RangeError(`Unsupported sync range: ${range}`);
+  }
+  return shiftDateByMonths(date, direction * months);
```
ridafkih: Just use a map here to determine the shift ..

PR #455, opened 2026-08-01
Add option to create synced events as Out of Office

### LINKED ISSUE #412: Add option to create synced events as "Out of Office" type in Google Calendar
Love Keeper! Would love to be able to configure that all synced events are created as "Out of Office" events in Google Calendar (the new type that auto-declines meetings). Currently, I can only exclude Out of Office events, but not create them.

Use case: Sync events from private calendar to work calendar as Out of Office blocks, so meetings get automatically declined during those times.

loeffelandi:
## Summary
- Adds a source sync setting **Create as Out of Office** that forces synced events to `availability: oof`
- Google destinations create timed events via `events.insert` with `eventType: outOfOffice` (import cannot create OOO), so conflicting meetings auto-decline
- All-day events fall back to normal busy blocks (Google OOO cannot be all-day)

Fixes #412

## Test plan
- [ ] Enable **Create as Out of Office** on a source that syncs to a Google primary calendar
- [ ] Confirm timed events appear as Out of Office and auto-decline conflicting invites
- [ ] Confirm all-day events still sync as regular busy events (no churn)
- [ ] Toggle the setting off and confirm subsequent syncs recreate normal events
- [ ] `bun test` in `packages/calendar` for Google destination serialize/provider tests

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/calendar/src/core/events/events.ts
+++ b/packages/calendar/src/core/events/events.ts
@@ -176,6 +176,7 @@ const getEventsForCalendarsWithDiagnostics = async (
       excludeEventName: calendarsTable.excludeEventName,
       excludeFocusTime: calendarsTable.excludeFocusTime,
       excludeOutOfOffice: calendarsTable.excludeOutOfOffice,
+      createAsOutOfOffice: calendarsTable.createAsOutOfOffice,
       availability: eventStatesTable.availability,
       description: eventStatesTable.description,
       endTime: eventStatesTable.endTime,
@@ -252,7 +253,9 @@ const getEventsForCalendarsWithDiagnostics = async (
       calendarId: result.calendarId,
       calendarName: result.calendarName,
       calendarUrl: result.calendarUrl,
-      availability: parseAvailability(result.availability),
+      availability: result.createAsOutOfOffice
+        ? "oof"
+        : parseAvailability(result.availability),
       description: excludeOrAbsent(result.excludeEventDescription, result.description),
       endTime: result.endTime,
       eventStateId: result.id,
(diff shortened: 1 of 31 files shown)


packages/calendar/src/core/events/events.ts
```diff
@@ -252,7 +253,9 @@ const getEventsForCalendarsWithDiagnostics = async (
       calendarId: result.calendarId,
       calendarName: result.calendarName,
       calendarUrl: result.calendarUrl,
-      availability: parseAvailability(result.availability),
+      availability: result.createAsOutOfOffice
+        ? "oof"
+        : parseAvailability(result.availability),
```
ridafkih: Would love to be able to extend this effort to create as busy/free/oof, etc.? Can we create an enum instead of just having this be a boolean piped through?

saranatour1: Enums, yuck!

loeffelandi:
Addressed the review feedback: replaced the boolean with a `forcedAvailability` string union (`busy` | `free` | `oof` | `null`) so destinations can be forced to any of those states, or left as the source value (`null` / “Source” in the UI).

Went with a string union rather than a TypeScript `enum` to match the existing availability typing in the codebase.

packages/calendar/src/core/events/events.ts
loeffelandi: Agreed on TypeScript `enum`s — we avoided those. This is a plain text column + string union (`'busy' | 'free' | 'oof' | null`), same pattern as the existing `EventAvailability` typing elsewhere in the codebase.

ridafkih: > Enums, yuck!

`} as const` for life.

PR #483, opened 2026-08-11
refactor: move the comparison article into the blog

### LINKED ISSUE #469: Add alternative and comparison landing pages
We have no pages targeting the "X alternative" or "X vs Y" searches people actually run when picking a calendar sync tool. That intent is entirely uncontested for us right now.

Suggest a `/compare/` route group, starting with the tools that come up most in our own issues and support threads, plus an open-source roundup. Honest comparisons only, including where we lose. Same sitemap caveat as /features and /pricing.

ridafkih:
Closes #469. Reworked after review: the comparison content is blog content, so the separate `/compare` collection is gone and the roundup now lives at `/blog/open-source-calendar-sync` alongside everything else. The blog MDX pipeline gained real custom components, so FAQs render as the same collapsibles the homepage uses instead of hand-rolled markdown headings.

- Deleted `plugins/compare.ts`, `lib/compare-articles.ts`, the `compare/` route group and `compareArticleSchema`; the article is a `BlogPosting` and `/compare` now 404s. `plugins/sitemap.ts` keeps the shared `discoverContentEntries` reader and the missing-frontmatter crash fix, but discovers blog content only
- `Prose` takes `allowedTags` and `components`, so blog MDX can use `<faq>` and `<faq-item question="...">`, backed by `MarketingFaq*` and `Collapsible`
- `plugins/blog.ts` extracts the FAQ entries out of the content at build time, so `FAQPage` JSON-LD stays derived from exactly what renders
- Dropped the sixteen per-source "Checked 11 August 2026" stamps; the one statement above the Sources list covers them
- Dropped the two asides volunteering that Keeper.sh has no end-to-end encryption. EteSync's own encryption is still described accurately in the table, its section and the sources list

### DIFF
--- a/applications/web/plugins/sitemap.ts
+++ b/applications/web/plugins/sitemap.ts
@@ -61,29 +61,35 @@ function buildBlogIndexEntry(blogEntries: SitemapEntry[]): SitemapEntry {
   return { loc: `${SITE_URL}/blog`, lastmod };
 }
 
-function parseFrontmatter(raw: string): Record<string, unknown> {
-  const [, match] = raw.match(FRONTMATTER_PATTERN);
-  if (!match) return {};
-  return parseYaml(match);
+function parseFrontmatter(raw: string, file: string): Record<string, unknown> {
+  const match = raw.match(FRONTMATTER_PATTERN);
+  if (!match) {
+    throw new Error(`"${file}" is missing a YAML frontmatter block.`);
+  }
+  return parseYaml(match[1]);
 }
 
-function discoverBlogEntries(blogDir: string): SitemapEntry[] {
-  const files = readdirSync(blogDir).filter((f) => f.endsWith(".mdx"));
+function discoverContentEntries(
+  directory: string,
+  basePath: string,
+  label: string,
+): SitemapEntry[] {
+  const files = readdirSync(directory).filter((file) => file.endsWith(".mdx"));
 
   return files.map((file) => {
-    const raw = readFileSync(join(blogDir, file), "utf-8");
-    const frontmatter = parseFrontmatter(raw);
+    const raw = readFileSync(join(directory, file), "utf-8");
+    const frontmatter = parseFrontmatter(raw, file);
 
     if (typeof frontmatter.slug !== "string") {
-      throw new Error(`Blog post "${file}" is missing a slug.`);
+      throw new Error(`${label} "${file}" is missing a slug.`);
     }
 
     if (typeof frontmatter.updatedAt !== "string") {
-      throw new Error(`Blog post "${file}" is missing updatedAt.`);
+      throw new Error(`${label} "${file}" is missing updatedAt.`);
     }
 
     return {
-      loc: `${SITE_URL}/blog/${frontmatter.slug}`,
+      loc: `${SITE_URL}${basePath}/${frontmatter.slug}`,
       lastmod: frontmatter.updatedAt.slice(0, 10),
     };
   });
@@ -124,7 +130,7 @@ export function sitemapPlugin(): Plugin {
     },
 
     generateBundle() {
-      const blogEntries = discoverBlogEntries(blogDir);
+      const blogEntries = discoverContentEntries(blogDir, "/blog", "Blog post");
       const entries = [
         ...readStaticEntries(pagesFile),
         buildBlogIndexEntry(blogEntries),
--- a/applications/web/plugins/blog.ts
+++ b/applications/web/plugins/blog.ts
@@ -23,10 +23,37 @@ const blogPostMetadataSchema = type({
 
 type BlogPostMetadata = typeof blogPostMetadataSchema.infer;
 
+interface BlogPostFaqEntry {
+  answer: string;
+  question: string;
+}
+
 export interface ProcessedBlogPost {
   content: string;
+  faq: BlogPostFaqEntry[];
   metadata: BlogPostMetadata;
   slug: string;
+}
+
+const FAQ_ITEM_PATTERN = /<faq-item question="([^"]*)">([\s\S]*?)<\/faq-item>/g;
+
+function extractFaqEntries(content: string, filePath: string): BlogPostFaqEntry[] {
+  const entries = [...content.matchAll(FAQ_ITEM_PATTERN)].map((match) => ({
+    answer: match[2].trim(),
+    question: match[1].trim(),
+  }));
+
+  const incomplete = entries.find(
+    (entry) => entry.question.length === 0 || entry.answer.length === 0,
+  );
+
+  if (incomplete) {
+    throw new Error(
+      `Blog post "${filePath}" has a faq-item with an empty question or answer.`,
+    );
+  }
+
+  return entries;
 }
 
 function toIsoDate(value: string): string {
@@ -197,7 +224,7 @@ export function processBlogDirectory(
     slugCounts.set(baseSlug, seenCount + 1);
     const slug = seenCount === 0 ? baseSlug : `${baseSlug}-${seenCount + 1}`;
 
-    return { content, metadata, slug };
+    return { content, faq: extractFaqEntries(content, file), metadata, slug };
   });
 
   return posts.sort((a, b) =>
--- a/applications/web/src/components/ui/primitives/markdown-component-map.ts
+++ b/applications/web/src/components/ui/primitives/markdown-component-map.ts
@@ -12,8 +12,11 @@ import {
   MarkdownParagraph,
   MarkdownRule,
   MarkdownTable,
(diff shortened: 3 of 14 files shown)


applications/web/plugins/sitemap.ts
```diff
@@ -17,6 +17,11 @@ const staticEntries: SitemapEntry[] = [
   { loc: `${SITE_URL}/blog`, lastmod: "2026-03-09" },
   { loc: `${SITE_URL}/privacy`, lastmod: "2025-12-01" },
   { loc: `${SITE_URL}/terms`, lastmod: "2025-12-01" },
+  { loc: `${SITE_URL}/compare`, lastmod: "2026-08-11" },
+  { loc: `${SITE_URL}/compare/open-source-calendar-sync`, lastmod: "2026-08-11" },
+  { loc: `${SITE_URL}/compare/onecal-alternative`, lastmod: "2026-08-11" },
+  { loc: `${SITE_URL}/compare/calendarbridge-alternative`, lastmod: "2026-08-11" },
+  { loc: `${SITE_URL}/compare/reclaim-alternative`, lastmod: "2026-08-11" },
```
ridafkih: Very nice.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+}) {
+  return (
+    <header className="flex flex-col gap-2">
+      <Text size="xs" tone="muted" className="uppercase opacity-75">
+        {eyebrow}
+      </Text>
+      <Heading1>{title}</Heading1>
+      <Text size="base" tone="muted" className="max-w-[68ch] leading-6">
+        {summary}
+      </Text>
```
ridafkih: We should probably standardize the text styles here so that we don't result in inconsistent styles across the code base. Right now another agent might do a different max width and a different leading, depending on the context, which is inconsistent.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+}: PropsWithChildren<{ title: string; description?: ReactNode }>) {
+  return (
+    <section className="flex flex-col gap-4">
+      <div className="flex flex-col gap-1.5">
+        <Heading2 as="h2">{title}</Heading2>
+        {description ? (
+          <Text size="sm" tone="muted" className="max-w-[68ch] leading-6">
+            {description}
+          </Text>
+        ) : null}
```
ridafkih: I would rather we spread than do a turn area here or just have it in its own separate component. Ditto on the twxt styles.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+      {children}
+    </section>
+  );
+}
+
+export function CompareProse({ children }: PropsWithChildren) {
+  return (
+    <div className="flex flex-col gap-3 max-w-[68ch]">
+      {children}
+    </div>
```
ridafkih: Why do we have pros in a div?

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+  return (
+    <sup className="ml-1 tabular-nums">
+      {positions.map((position, index) => (
+        <span key={sources[position].id}>
+          {index > 0 ? <span className="text-foreground-disabled">,</span> : null}
+          <a
+            href={`#source-${sources[position].id}`}
+            className="text-foreground-muted hover:text-foreground underline underline-offset-2"
+          >
+            {position + 1}
```
ridafkih: Anchor tag styles here should be consistent with the rest of the application.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+        <thead>
+          <tr className="border-b border-interactive-border">
+            <th scope="col" className="w-1/4 px-4 py-3 font-medium text-foreground">
+              Dimension
+            </th>
+            <th scope="col" className="px-4 py-3 font-medium text-foreground">
+              Keeper.sh
+            </th>
+            <th scope="col" className="px-4 py-3 font-medium text-foreground">
+              {rivalName}
```
ridafkih: A table is probably a fine pattern here. We should be using grid and overwriting the styles of the actual table itself. I'm questioning the value of a comparison table versus a properly laid out MDX document that can be featured as a page.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+            <tr key={row.dimension} className="border-b border-interactive-border last:border-b-0">
+              <th scope="row" className="px-4 py-3 align-top font-normal text-foreground">
+                {row.dimension}
+              </th>
+              <td className="px-4 py-3 align-top text-foreground-muted">{row.keeper}</td>
+              <td className="px-4 py-3 align-top text-foreground-muted">
+                {row.rival}
+                {row.sourceIds ? (
+                  <CompareCitation sourceIds={row.sourceIds} sources={sources} />
+                ) : null}
```
ridafkih: I don't think these ternaries are good and I don't think they'll fly with the linter.

applications/web/src/features/compare/components/compare-page.tsx
```diff
@@ -0,0 +1,238 @@
...
+            align="left"
+            href={href}
+            rel="noopener noreferrer nofollow"
+            size="xs"
+            target="_blank"
+            tone="muted"
+          >
+            {href.replace(/^https?:\/\//, "")}
+          </ExternalTextLink>
+          <CompareCitation sourceIds={sourceIds} sources={sources} />
```
ridafkih: We should be using a URL constructor to deconstruct the URL rather than using replacement. That way we can handle edge cases with different protocols and different URL formats.

applications/web/src/lib/compare.ts
```diff
@@ -0,0 +1,47 @@
...
+  {
+    path: "/compare/open-source-calendar-sync",
+    title: "Open-source calendar sync tools",
+    blurb:
+      "The self-hostable projects people reach for when they want calendars in sync, sorted by what they actually do. Most of them are CalDAV servers or clients rather than sync tools.",
+  },
+  {
+    path: "/compare/onecal-alternative",
+    title: "Keeper.sh vs OneCal",
+    blurb:
```
ridafkih: Yeah, I definitely think that rather than this, we should do dedicated MDX articles. I think most of the work in this PR is gonna be completely redone. We should also do more research and include more competitors. There's more players in the space.

applications/web/src/routes/(marketing)/compare/calendarbridge-alternative.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+const SOURCES: CompareSource[] = [
+  {
+    id: "cb-home",
+    label: "CalendarBridge homepage — providers, sync description, scheduling links, busy-only privacy, trial wording.",
+    url: "https://calendarbridge.com",
+    checkedAt: CHECKED_AT,
+  },
+  {
+    id: "cb-pricing",
```
ridafkih: A big reason of why we're doing this is to counteract incorrect information and lies that other competitors are putting in their comparison pages. We need to make sure that we're offering competitors the same degree of honesty that we hope that they provide us. With that said, we should do more research to ensure that we're actually capturing the feature set that they support. Make sure that we're not doing anything dishonest in any of these NDX articles that we're gonna end up writing. 

Remember that each product has its actual pros and cons. We're not just trying to sell people keeper because. While we shouldn't lean too hard on pushing people away from Keeper, users should be able to read between the lines and come out with an understanding of why another product is a better option for them if it actually is.

applications/web/src/routes/(marketing)/compare/calendarbridge-alternative.tsx
```diff
@@ -0,0 +1,298 @@
...
+  {
+    dimension: "Event detail copied",
+    keeper: "Title, description, location, times, timezone and recurrence. Attendees, organisers, conference links and reminders are never copied.",
+    rival: "Per-field toggles for subject, attendees, description, location, conference link and reminders, with time and date always synced.",
+    sourceIds: ["cb-manage-syncs"],
+  },
+  {
+    dimension: "Hiding detail",
+    keeper: "Per-calendar toggles to strip title, description and location, plus a title template. Synced events are not marked private on the destination.",
+    rival: "A free/busy-only mode that writes events titled Busy, and an All Private toggle that marks every synced event private on the destination.",
```
ridafkih: This file is slop and unmaintainable. MDX documents please.

Issue #460, opened 2026-08-11
Self-Hosted MCP Documentation Question


afsabresfan:
I'm trying to understand what the following information means in the self-hosted MCP section:

1. Run the keeper-mcp container with MCP_PORT, MCP_PUBLIC_URL, DATABASE_URL, BETTER_AUTH_SECRET, and BETTER_AUTH_URL.
2. Set MCP_PUBLIC_URL on the api service to the same value (e.g. https://keeper.example.com/mcp).
3. Set VITE_MCP_URL on the web service to the internal URL of the MCP container (e.g. http://mcp:3002).

I'm using the standalone method of docker deployment. I _think_ from what I read that I should just add this after the keeper-standalone section in the docker compose:

```
  keeper:
    image: ghcr.io/ridafkih/keeper-standalone:2

  mcp:
    image: ghcr.io/ridafkih/keeper-mcp:latest
    environment:
        MCP_PORT: 4081
        MCP_PUBLIC_URL: http://192.168.1.198:4081/mcp
        DATABASE_URL: postgres://keeper:keeper@postgres:5432/keeper
        BETTER_AUTH_SECRET:
        BETTER_AUTH_URL: 
    ports:
      - "4081:3000"
    depends_on:
      api:
        condition: service_started
```

Is that correct? What do I use for BETTER_AUTH_SECRET and BETTER_AUTH_URL? And the MCP port should be different than the standard port, correct?

Thanks!!

ridafkih:
@afsabresfan You can ignore Claude's slop issue/PR attached, this is a complete oversight on my part. MCP should be built into the standalone image. Adding that now.

ridafkih:
@afsabresfan You should now be able to use this, make sure to clear your image cache if you have one - just follow the updated instructions in the README.md for the standalone image. c:

PR #496, opened 2026-08-12
docs: add MCP setup guidance for keeper-standalone

### LINKED ISSUE #485: Document MCP setup for keeper-standalone
The self-hosted MCP instructions only cover the individual service images and give no guidance for `keeper-standalone`, whose Postgres binds to localhost and is unreachable from a sibling MCP container.

Raised in #460.

ridafkih:
The self-hosted MCP instructions only worked for deployments where the MCP container can reach Postgres over the Compose network, which standalone can't do — its Postgres binds to `localhost` and only port 80 is published, so a sibling `keeper-mcp` container has no route to the database. This documents sharing standalone's network namespace with `network_mode: "service:keeper"`, and splits the section so it's clear which instructions apply to which image. Closes #485, raised in #460.

- Verified end to end by actually running the documented compose: MCP container starts, queries standalone's Postgres over loopback, `/mcp` is proxied by the bundled web service via `VITE_MCP_URL`, and an unauthenticated `initialize` returns `401` with `WWW-Authenticate: Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource"`.
- Confirmed the negative cases too: a sibling container on the Compose network never connects to `keeper:5432`, and with `VITE_MCP_URL` unset `/mcp` returns `404` from the SPA.
- Answers the questions from #460 directly — `BETTER_AUTH_URL` is the instance base URL rather than the MCP URL, `BETTER_AUTH_SECRET` must match across both containers, and no second published port is needed since the proxy serves `/mcp` on the existing one.
- Heads up, separate from this PR: `ghcr.io/ridafkih/keeper-standalone:2` and `keeper-mcp:2` are not published — the registry only has `{{version}}`, `{{major}}.{{minor}}`, and `latest`, since `docker-publish.yml` has no `type=semver,pattern={{major}}`. I kept `:2` here to match the rest of the README, but I verified against `2.13`, and every existing self-hosting example has the same problem.
- Also worth a look separately: setting `VITE_MCP_URL` to an empty string (rather than leaving it unset) fails env validation and takes the whole web service down with a 502 on every route.

### DIFF
--- a/README.md
+++ b/README.md
@@ -612,13 +612,93 @@ Example Claude Code MCP configuration:
 >
 > MCP is fully optional. All MCP-related environment variables are optional across every service and image. If they are not set, Keeper starts normally without MCP functionality. Existing self-hosted deployments are unaffected.
 
-The MCP server is proxied through the web service at `/mcp`, the same way the API is proxied at `/api`. MCP is **not** bundled in the `keeper-standalone` or `keeper-services` convenience images — run the `keeper-mcp` image as a separate container alongside them.
+The MCP server is proxied through the web service at `/mcp`, the same way the API is proxied at `/api`. MCP is **not** bundled in the `keeper-standalone` or `keeper-services` convenience images — run the `keeper-mcp` image as a separate container alongside them. How you wire that container up depends on which image you started from, so follow the section that matches your deployment.
 
-To enable MCP on a self-hosted instance:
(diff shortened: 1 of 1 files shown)


ridafkih:
Superseded by #540, which bundles MCP into the standalone image so the sidecar this documents is no longer needed.

PR #507, opened 2026-08-12
feat: encrypt OAuth access and refresh tokens at rest

ridafkih:
`ENCRYPTION_KEY` becomes required in api, cron and worker, and this PR is steps 1 and 2 only — dual-read plus encrypt-on-write, with the backfill (`bun run backfill:oauth-tokens`) and the later removal of plaintext-read support as separate steps.


ridafkih:
I built this independently before spotting this PR — same design, and yours is better staged, so I've binned mine and closed #555 as a duplicate. Two things from that attempt might be worth stealing.

**1. Make the compiler prove the call-site sweep.** Rather than editing each read and write, type the columns with a Drizzle `customType` whose TS type is an opaque branded value. The SQL type stays `text`, so no column migration:

```ts
declare const storedTokenBrand: unique symbol;
interface StoredToken { readonly [storedTokenBrand]: true; readonly stored: string }

const storedToken = customType<{ data: StoredToken; driverData: string }>({
  dataType: () => "text",
  fromDriver: (value) => readStoredToken(value),
  toDriver: (value) => storedTokenValue(value),
});
```

Every site that wants a `string` then fails to compile unless it goes through `encryptToken`/`decryptToken`, so `bun run types` becomes the evidence that nothing was missed. With 11 files referencing `oauthCredentials` and ~300 `accessToken`/`refreshToken` references in the monorepo, a grep-based sweep is hard to trust; this one can't silently miss a path. It also makes the later "retire plaintext reads" PR safe to land, since anything still handling a raw string will refuse to build.

**2. Assert on the whole write, not the field.** My first version of the regression test asserted `expect(write.accessToken).not.toBe(PLAINTEXT)`, which passes if the token leaks into any other column. Serialising the entire write is stricter and cheap:

```ts
expect(JSON.stringify(write)).not.toContain(PLAINTEXT_ACCESS_TOKEN);
expect(JSON.stringify(write)).not.toContain(PLAINTEXT_REFRESH_TOKEN);
expect(decryptToken(write.accessToken, ENCRYPTION_KEY)).toBe(PLAINTEXT_ACCESS_TOKEN);
```

Paired with one that asserts a keyless refresh throws and performs **zero** writes, rather than only that it throws.

Unrelated, on the `--verify` step: worth having it fail on any row whose ciphertext will not decrypt with the current key, not only on rows still in plaintext. A rotated or mistyped `ENCRYPTION_KEY` produces rows that are encrypted, so `remaining: 0`, but unreadable — which is the one failure mode that silently passes the gate.

PR #514, opened 2026-08-12
refactor: make a keyless configuration unrepresentable

ridafkih:
Must merge AFTER #507 — merging it first would break keyless deployments that #507 has not yet made impossible, since it makes `ENCRYPTION_KEY` required without #507's startup migration notice. #507 makes the key required at runtime; this removes the now-unreachable branches that handled its absence, so nobody can later "restore" keyless operation and silently reintroduce credential loss. The optionality all originates in the three env schemas, so those are tightened first and the guards then fail to typecheck rather than merely being unused.

- `services/{api,cron,worker}/src/env.ts`: `ENCRYPTION_KEY` `"string?"` → `"string"`
- `services/api/src/utils/caldav.ts` and `utils/caldav-sources.ts`: dropped the two "key not configured" throws, and the `resolvedEncryptionKey` narrowing alias with them
- `services/api/src/mutations/index.ts`: `MutationDependencies.encryptionKey` now required; dropped the four `deps.encryptionKey &&` conjuncts on the CalDAV dispatch path plus one on pending invites
- `services/api/src/read-models.ts`: `KeeperApiOptions.encryptionKey` and the `options` argument now required, so the seven read-only `createKeeperApi(database)` routes pass it through
- `packages/sync/src/{sync-user,resolve-provider}.ts`: `SyncConfig.encryptionKey` and `ResolveProviderOptions.encryptionKey` now required; dropped the `&& options.encryptionKey` conjunct
- `services/cron/src/jobs/ingest-sources.ts`: dropped the early return that skipped CalDAV ingest entirely when the key was unset

Every removed conjunct sat beside conditions that still gate the branch, so branch selection is unchanged once the key is guaranteed. #507 independently tightens the env schemas, `MutationDependencies`, `KeeperApiOptions` and `SyncConfig`; those lines will conflict and should resolve in favour of #507. README and `.env.template` updates are left to #507. `services/mcp` and `applications/web` untouched.

`bun run types`, `bun run lint` and `bun run unused` pass. `bun run test` passes except the pre-existing host-ICU-dependent Morocco VTIMEZONE test in `packages/calendar`, which this PR does not touch.

### DIFF
--- a/packages/sync/src/resolve-provider.ts
+++ b/packages/sync/src/resolve-provider.ts
@@ -158,7 +158,7 @@ interface ResolveProviderOptions {
   userId: string;
   accountId: string;
   oauthConfig: OAuthConfig;
-  encryptionKey?: string;
+  encryptionKey: string;
   refreshLockStore?: RefreshLockStore | null;
   rateLimiter?: RedisRateLimiter;
   signal?: AbortSignal;
@@ -179,7 +179,7 @@ const resolveSyncProvider = (options: ResolveProviderOptions): Promise<CalendarS
     );
   }
 
-  if (CALDAV_PROVIDERS.has(options.provider) && options.encryptionKey) {
+  if (CALDAV_PROVIDERS.has(options.provider)) {
     return resolveCalDAVProvider(
       options.database,
       options.calendarId,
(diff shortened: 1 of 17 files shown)


ridafkih:
Folded into #507. These guards only become unreachable because #507 makes `ENCRYPTION_KEY` required and adds the boot-time check that exits on any falsy value, so the two changes are inseparable — landing this alone would remove guards that are still load-bearing. Applied there instead. Leaving the branch in place until #507 lands.

PR #534, opened 2026-08-12
fix(web): stop markdown table columns collapsing to word width

ridafkih:
The comparison table in `/blog/best-calendar-sync-tools` had no minimum column width, so auto table layout squeezed every column down to its longest word — "Run it yourself" rendered 75px wide and wrapped to one word per line. Worse, the `overflow-x-auto` wrapper propagated the table's min-content width straight up into the `minmax(auto,48rem)` content track in `layout.tsx`, so at 375px the page itself grew to 784px behind an `overflow-x: hidden` body: the right half of the table was unreachable rather than scrollable. Cells now carry a real minimum and the wrapper is sized from its container instead of its contents, so the table scrolls inside its own box.

- `min-w-32` on every cell, `first:min-w-48` to keep the feature-name column wider
- `[contain:inline-size]` on the scroll wrapper so its min-content no longer feeds the page grid
- Column widths before, at both 1440px and 375px: 159 / 75 / 131 / 102 / 87 / 94 / 100 / 105 px. After: 192 / 128 / 131 / 128 / 128 / 128 / 128 / 128 px
- `document.body.scrollWidth` at a 375px viewport: 784px before, 375px after; the wrapper now scrolls 343px visible over 1092px of table

https://claude.ai/code/session_01NmFcV1wCy2785Spmy5dEzY

### DIFF
--- a/applications/web/src/components/ui/primitives/markdown-components.tsx
+++ b/applications/web/src/components/ui/primitives/markdown-components.tsx
@@ -1,4 +1,5 @@
-import type { JSX } from "react";
+import { isValidElement, type JSX, type ReactNode } from "react";
+import { cn } from "@/utils/cn";
 import { Heading1, Heading2, Heading3 } from "./heading";
 import { ListItem, OrderedList, UnorderedList } from "./list";
 import { Text } from "./text";
@@ -8,8 +9,30 @@ type MarkdownElementProps<Tag extends keyof JSX.IntrinsicElements> =
   node?: unknown;
 };
 
+const PROSE_CELL_CHARACTER_THRESHOLD = 20;
+
 function isExternalHttpLink(href: string): boolean {
   return href.startsWith("http://") || href.startsWith("https://");
+}
+
+function countTextCharacters(node: ReactNode): number {
+  if (typeof node === "string" || typeof node === "number") {
+    return String(node).length;
+  }
+
+  if (Array.isArray(node)) {
+    return node.reduce<number>((total, child) => total + countTextCharacters(child), 0);
+  }
+
+  if (isValidElement<{ children?: ReactNode }>(node)) {
+    return countTextCharacters(node.props.children);
+  }
+
+  return 0;
+}
+
+function isProseCell(children: ReactNode): boolean {
+  return countTextCharacters(children) > PROSE_CELL_CHARACTER_THRESHOLD;
 }
 
 export function MarkdownHeadingOne({ children }: MarkdownElementProps<"h1">) {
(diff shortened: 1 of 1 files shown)


applications/web/src/components/ui/primitives/markdown-components.tsx
```diff
@@ -105,7 +105,7 @@ export function MarkdownTable({
   children,
 }: MarkdownElementProps<"table">) {
   return (
-    <div className="my-4 overflow-x-auto">
+    <div className="my-4 overflow-x-auto [contain:inline-size]">
```
ridafkih: Not needed.

applications/web/src/components/ui/primitives/markdown-components.tsx
```diff
@@ -116,11 +116,11 @@ export function MarkdownTable({
 export function MarkdownTableHeader({
   children,
 }: MarkdownElementProps<"th">) {
-  return <th className="border border-interactive-border px-2 py-1 text-left font-medium text-foreground">{children}</th>;
+  return <th className="min-w-32 border border-interactive-border px-2 py-1 text-left font-medium text-foreground first:min-w-48">{children}</th>;
```
ridafkih: We should be selectively applying this to mid-form text content, not all cells.

applications/web/src/components/ui/primitives/markdown-components.tsx
```diff
@@ -116,11 +116,11 @@ export function MarkdownTable({
...
 }
 
 export function MarkdownTableCell({
   children,
 }: MarkdownElementProps<"td">) {
-  return <td className="border border-interactive-border px-2 py-1">{children}</td>;
+  return <td className="min-w-32 border border-interactive-border px-2 py-1 first:min-w-48">{children}</td>;
```
ridafkih: Ditto.

applications/web/src/components/ui/primitives/markdown-components.tsx
ridafkih: Done — the floor is now content-aware rather than blanket. `isProseCell` walks the cell's children, and only cells carrying more than 20 characters of text get `min-w-32`; value cells ("Yes", "No", "$5/month", a dash) get nothing and size to their content. On the comparison post that floors 44 of 80 cells.

On that table it costs nothing and saves a little: the table goes 1092px → 1059px because the `Tool` column now sits at its natural 159px instead of the 192px the old `first:min-w-48` forced. I dropped that first-column floor too — it was the same blanket padding, and the column is already the widest non-prose one on its own.

The win shows up on a table that actually has short columns. Five columns, four of them values plus one prose column, measured at 375px:

- blanket floor: 128 / 128 / 128 / 128 / 128, table 641px
- selective floor: 72 / 87 / 70 / 72 / 128, table 429px

33% less horizontal scroll, and the prose column keeps its floor. At 1440px the same table gives the prose column 365px instead of 255px.

Comparison post after the change, identical at 1440px and 375px: 159 / 128 / 131 / 128 / 128 / 128 / 128 / 128, table 1059px, wrapper 768px (desktop) and 343px (mobile) client over 1059px scroll. `document.body.scrollWidth` is 1440 at 1440 and 375 at 375 — no page-level overflow reintroduced.

applications/web/src/components/ui/primitives/markdown-components.tsx
ridafkih: Same change on `td` — `isProseCell(children) && "min-w-32"` instead of the unconditional class, and the `first:min-w-48` is gone. Numbers in the reply on the `th` comment above.

applications/web/src/components/ui/primitives/markdown-components.tsx
ridafkih: Understood that this one is withdrawn, so it stays — noting why for the record, since it is doing real work.

Without it the wrapper's min-content propagates into the `minmax(auto,48rem)` content track in `layout.tsx`: `overflow` zeroes a flex/grid item's *automatic minimum size*, not its intrinsic contribution. On `main` today that makes `main` 768px wide at a 375px viewport and `document.body.scrollWidth` 784px — and because `html`/`body` are `overflow-x: hidden` it clips rather than scrolls, so the right side of the table is unreachable on a phone. With it, `body.scrollWidth` is 375px and the wrapper scrolls 343px over 1059px on its own.

`w-0 min-w-full` and `display: grid` on the wrapper both measured identically if you would rather not have the `contain` property in there — happy to swap.

PR #545, opened 2026-08-12
fix(caldav): chunk calendar-multiget and reject short responses (#461)

### LINKED ISSUE #461: CalDAV ingestion silently drops events when calendar-multiget response is capped (Zoho: 1,000)
### Summary

CalDAV source ingestion can silently miss native events when a server limits the number of objects returned by a single `calendar-multiget` response.

With Zoho Calendar, `calendar-query` returned 1,971 matching `.ics` hrefs, but the subsequent single multiget returned only 1,000 `calendar-data` entries with a successful HTTP 207 response. If native events fall outside that first response set, Keeper reports a successful ingestion with zero events added.

### Environment

- Keeper 2.13.5 (`keeper-services`, ARM64, self-hosted)
- Zoho Calendar via CalDAV (`calendar.zoho.eu`)
- Calendar configured as both a source and a destination

### Reproduction and diagnostics

1. Configure a Zoho CalDAV calendar containing more than 1,000 events in Keepers sync window.
2. Configure that calendar as a source for another destination.
3. Allow Keeper-generated events to accumulate in the Zoho destination.
4. Observe that CalDAV source ingestion succeeds every minute but adds zero native events.

Direct CalDAV diagnostics using the same account and sync window showed:

- `calendar-query`: 1,971 `.ics` hrefs
- One multiget containing all hrefs: exactly 1,000 objects returned
- All first 1,000 UIDs ended with `@keeper.sh`, so the source adapter correctly filtered them as Keeper-created events
- Fetching the hrefs in batches of 500 returned:
  - offset 0: 500 Keeper events, 0 native
  - offset 500: 500 Keeper events, 0 native
  - offset 1,000: 497 Keeper events, 1 native
  - offset 1,500: 471 Keeper events, 0 native

The native event was therefore present and readable through CalDAV, but it was never included in Keepers single capped multiget response. Keeper logged the ingestion as successful with `events_added=0`, and no source event state was created.

### Expected behavior

Keeper should retrieve and process all hrefs returned by `calendar-query`, including native events beyond a providers per-response limit.

### Actual behavior

Keeper sends one multiget for the entire href list. Zoho silently caps the successful response at 1,000 objects, so remaining events are ignored and cannot be pushed to mapped destination calendars.

### Suspected location

`CalDAVClient.fetchCalendarObjects()` gathers all object URLs and passes the complete list to one `calendarMultiGet()` call:

https://github.com/ridafkih/keeper.sh/blob/1c274dbe74fce3b8464c8686e1cec63c14e34557/packages/calendar/src/providers/caldav/shared/client.ts

The same behavior is present in 2.13.5 and current `main`.

### Suggested fix

Chunk `calendarObjectUrls` into bounded batches (for example, 250-500), issue one multiget per batch with limited concurrency, and combine the responses. It may also be useful to detect and report hrefs missing from otherwise successful multiget responses.

Reducing the destination below 1,000 objects is only a temporary workaround because the issue returns as the calendar grows.

ridafkih:
`CalDAVClient.fetchCalendarObjects` handed the entire href list to a single `calendar-multiget`, and Zoho silently caps that response at 1,000 objects with a successful 207 — so events past the cap were dropped and ingestion logged `events_added=0` as a success. It now runs `calendarQuery` for hrefs, chunks at 250, issues one multiget per batch, and reconciles requested against returned hrefs, throwing rather than returning a short set. Signature is unchanged, so all three call sites are fixed without edits.

- Sequential batches, not concurrent: `mapAuthenticationFailure` reads a single mutable `lastResponseStatus`, so parallel requests would let a 207 mask a 401
- `CalDAVIncompleteMultiGetError` deliberately carries no status field, so `isCalDAVAuthenticationError` won't flip `needsReauthentication` on a live account
- `chunkArray` lifted out of `providers/google/shared/batch.ts` into `core/utils` rather than duplicated
- New cron slug `provider-response-incomplete` so truncation is visible in Grafana instead of reading as a clean sync
- 7 new tests fail on `main` for the right reason, including the reporter's exact case (1,001 hrefs, one native event past the cap → `expected [] to have a length of 1`)

Batch size is a module constant matching `GOOGLE_BATCH_MAX_SIZE`; happy to make it per-source configurable if a server needs it.

Closes #461

https://claude.ai/code/session_016bBKPZhpDah8CHTkb8srRD


packages/calendar/tests/providers/caldav/shared/caldav-server-stub.ts
```diff
@@ -0,0 +1,139 @@
+const SERVER_URL = "https://caldav.example.com";
```
ridafkih: This is not how we do test fixtures/utils in this project.

packages/calendar/src/providers/caldav/shared/multiget.ts
```diff
@@ -0,0 +1,136 @@
+const CALDAV_MULTIGET_BATCH_SIZE = 250;
```
ridafkih: We need to do research on proper `multiget` handling in NodeJS, because this file is one of my least favourite files I've ever seen.

packages/calendar/src/providers/caldav/shared/client.ts
```diff
@@ -207,10 +217,47 @@ class CalDAVClient {
...
+      const batches = chunkArray(requestedPathnames, CALDAV_MULTIGET_BATCH_SIZE);
+      const returnedObjects: CalendarObject[] = [];
+
+      for (const batch of batches) {
+        const objects = await client.fetchCalendarObjects({
+          calendar: { url: params.calendarUrl },
+          objectUrls: batch,
+        });
+        returnedObjects.push(...objects.filter((object) => hasCalendarData(object)));
+      }
```
ridafkih: We can parallelize these, no?

PR #552, opened 2026-08-12
feat(ical): support multiple named iCal feeds per user

ridafkih:
Users now get more than one iCal link. Each feed carries its own opaque token, calendar selection, privacy toggles, and event filters, so a "work only" link can be shared without exposing everything else. Free stays at one feed, Pro is unlimited, enforced at creation time through the existing `{ current, limit }` entitlements shape.

- Back-compat is the load-bearing constraint: `/api/cal/{username}.ics` keeps working and returning the same content. Migration `0079` backfills every existing user a default feed carrying their current settings and implicit calendar selection, and a test pins the legacy URL's output.
- New feeds use CSPRNG opaque tokens rather than the username, which was guessable and can't be shared across N feeds.
- No downgrade handling, per your call — the limit is enforced on create only, so a Pro user who downgrades keeps the feeds they already made.
- An adversarial pass caught the disclosure line claiming "event times only" whenever names were off, while the ICS still carried descriptions and locations; it now names each shared field independently.

Migration collision: `feat/refresh-calendars` also claims `0079`. Whichever merges second needs renumbering to `0080` with a regenerated snapshot.

https://claude.ai/code/session_016bBKPZhpDah8CHTkb8srRD


applications/web/src/routes/(dashboard)/dashboard/ical/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+function FeedCopyButton({ feed }: { feed: IcalFeed }) {
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = async (event: React.MouseEvent) => {
+    event.stopPropagation();
+    await navigator.clipboard.writeText(feed.icalUrl);
+    track(ANALYTICS_EVENTS.ical_link_copied);
+    setCopied(true);
+    setTimeout(() => setCopied(false), COPIED_RESET_MS);
```
ridafkih: This needs to be stored in a ref, and cancelled if you spam - otherwise it'll be super finniky.

applications/web/src/routes/(dashboard)/dashboard/ical/$feedId.tsx
```diff
@@ -0,0 +1,387 @@
...
+    <Collapsible
+      className="rounded-2xl border border-border-elevated bg-background-elevated"
+      trigger={<Text size="sm" tone="muted">Legacy link (still active)</Text>}
+    >
+      <div className="flex flex-col gap-2">
+        <Input readOnly value={legacyIcalUrl} className="text-sm" />
+        <Text size="sm" tone="muted">
+          Your original link keeps working, so anything already subscribed to it stays up to date.
+        </Text>
+      </div>
```
ridafkih: I think having a specific "Legacy" link is a little moronic, would be much nicer and more logical to simply fold this into the new primitive rather than producing a branch for no benefit.

applications/web/src/routes/(dashboard)/dashboard/ical/$feedId.tsx
```diff
@@ -0,0 +1,387 @@
...
+    return (
+      <div className="flex flex-col gap-1.5">
+        <BackButton fallback="/dashboard/ical" />
+      </div>
+    );
+  }
+
+  const patchFeed = (patch: Record<string, unknown>) => {
+    const swrKey = `/api/ical/feeds/${feedId}`;
+    mutate({ ...feed, ...patch } as IcalFeed, { revalidate: false });
```
ridafkih, line 117: Validation. We have it. Use it.

PR #592, opened 2026-08-12
fix(database): declare the live constraints in schema metadata

### LINKED ISSUE #578: schema.ts does not describe the constraints migrate.ts applies
`packages/database/scripts/migrate.ts` imperatively rewrites `event_mappings.eventStateId`'s FK to `ON DELETE SET NULL`, creates `event_mappings_source_calendar_idx` concurrently, and adds three CHECK constraints — none of which appear in `schema.ts` or the drizzle snapshot, which still declare `cascade`.

This is deliberate: 0077 is already applied in environments, so the journal cannot be edited. But it means `schema.ts` no longer describes live constraint behaviour, `drizzle-kit check` will not catch drift, and anyone running `drizzle-kit migrate`/`push` instead of the wrapper script gets bare columns with none of the integrity guarantees.

Worth either documenting the divergence prominently at both sites, or folding the constraints into a future migration so the metadata is honest again.

ridafkih:
Chose to fold the constraints into a new migration rather than document the divergence, because documentation does not fix the two real failures: `drizzle-kit check` still cannot see drift, and anyone running `drizzle-kit migrate`/`push` instead of the wrapper still gets a table with no integrity guarantees at all. 0077 stays untouched — 0079 is a new, fully guarded migration, so the journal history already applied in environments is unchanged. Verified `bun run generate` reports "No schema changes".

- `schema.ts` now declares `eventStateId` as `ON DELETE set null`, the `event_mappings_source_calendar_idx` index, and the three CHECK constraints (`event_mappings_identity_check`, `calendars_sync_ranges_check`, `calendars_ingest_coverage_check`), with the range list derived from `SYNC_RANGE_DEFINITIONS` so it cannot drift from `migrate.ts`.
- 0079 wraps every statement in an existence guard (`pg_constraint` lookups, `CREATE INDEX IF NOT EXISTS`) and adds all constraints `NOT VALID`, so it is a no-op on the databases where `migrate.ts` already applied them and never scans a large table inside the migration transaction.
- `migrate.ts` is unchanged: it still creates the index `CONCURRENTLY` on legacy databases and still runs `VALIDATE CONSTRAINT`, so 0079's `NOT VALID` additions are promoted to validated on the wrapper path.
- Verified against local Postgres 17: fresh database ends with all four constraints validated and the index present; a database migrated to 0078 with the old code then upgraded through 0079 stays validated with no downgrade; a bare `drizzle-kit migrate` (no wrapper) now produces all four constraints where it previously produced none.

Closes #578

https://claude.ai/code/session_01WTqvCoAV7hZzybpqV6cSj3

### DIFF
--- a/packages/database/drizzle/0080_parallel_jack_power.sql
+++ b/packages/database/drizzle/0080_parallel_jack_power.sql
@@ -1,0 +1,24 @@
+DO $$ BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'event_mappings_eventStateId_event_states_id_fk'
+      AND conrelid = 'event_mappings'::regclass
+      AND confdeltype = 'n'
+  ) THEN
+    ALTER TABLE "event_mappings"
+      DROP CONSTRAINT IF EXISTS "event_mappings_eventStateId_event_states_id_fk",
+      ADD CONSTRAINT "event_mappings_eventStateId_event_states_id_fk"
+        FOREIGN KEY ("eventStateId") REFERENCES "public"."event_states"("id")
+        ON DELETE set null ON UPDATE no action NOT VALID;
+  END IF;
+END $$;--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "event_mappings_source_calendar_idx" ON "event_mappings" USING btree ("sourceCalendarId");--> statement-breakpoint
+DO $$ BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_constraint
+    WHERE conname = 'event_mappings_identity_check'
+      AND conrelid = 'event_mappings'::regclass
+  ) THEN
+    ALTER TABLE "event_mappings" ADD CONSTRAINT "event_mappings_identity_check" CHECK ("event_mappings"."eventStateId" is not null or "event_mappings"."syncEventId" is not null) NOT VALID;
+  END IF;
+END $$;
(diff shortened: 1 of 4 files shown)


ridafkih:
Found while stacking this on #598 so 0079 would be exercised by the new migration job. It surfaced a defect that flat merging would have hidden.

**0079 breaks `normalizeCalendarCoverage` on databases with legacy invalid rows.** `migrate()` (`scripts/migrate.ts:425`) now applies 0079 and adds `calendars_ingest_coverage_check` as `NOT VALID`. `normalizeCalendarCoverage()` runs afterwards at :437 and repairs rows with five separate `UPDATE` statements. `NOT VALID` skips pre-existing rows but enforces on any row you write, so the first UPDATE — which fixes only `syncHistoricRange` — rewrites a row whose `ingestFutureRange` and window columns are still invalid, and Postgres rejects it:

```
error: new row for relation "calendars" violates check constraint "calendars_ingest_coverage_check"
Migration runner exited 1
```

Reproduced against real postgres via #598's `normalises out-of-range coverage before validating the coverage checks`. On `main` this cannot happen, because the constraint is added by `installEventMappingConstraints()` *after* normalization; this PR inverts that ordering by moving creation into the migration.

Affects any database carrying the legacy rows `normalizeCalendarCoverage` exists to repair, and it crash-loops under `set -e`.

Two directions, both yours to pick: run normalization before `migrate()`, or make the repair one atomic UPDATE that lands a fully valid row.

Separately, #598's test needs its seed to happen before the constraint exists (it currently applies all migrations, then seeds). I have that change locally and can push it with whichever fix you choose.

To be clear on process: 0079 itself is properly CLI-generated — machine journal entry, drizzle-assigned tag, and `bun run generate` reports "No schema changes". Only the SQL body was hand-edited, and I verified those guards are necessary: a CLI-pure body fails with `constraint "calendars_sync_ranges_check" for relation "calendars" already exists` on any database `migrate.ts` has already touched.

Issue #619, opened 2026-08-13
Polar 429s turn /api/auth/customer/state into a 500 and downgrade paying users in the dashboard


ridafkih:
`GET /api/auth/customer/state` returned HTTP 500 46 times in production over 08-11 → 08-13 (92 wide events, but each request is logged twice — once by `keeper-web` proxying, once by `keeper-api`). Every one of them has the same cause, visible only in better-auth's raw stderr:

```
ERROR [Better Auth]: Polar subscriptions list failed. Error: API error occurred: Status 429 Content-Type "". Body: ""
```

45 of those lines over the window, plus one `ResourceNotFound`. Polar rate-limits the org; the `@polar-sh/better-auth` plugin has no 429 handling, so the endpoint 500s.

### Not a deploy artifact

The 09:00 UTC cluster is a Loki bucketing artifact — the requests are actually at **08:00:56 → 08:01:41**, 31 API 500s in 45 seconds, all from one Chrome session on container `b5ffb3712c06`, which had been serving for ~11 minutes and served 200s afterwards. The other clusters land on long-lived containers hours into their life (`836b2c1526b6`, up 08:50→20:00, 500s at 11:01 and 12:02). So: not cold start, not a migration, not tied to any container's age.

What the deploy boundary correlation actually reflects is a **retry amplification loop**: the endpoint 500s → SWR/route-loader retries → each retry is another Polar call → the rate limit stays tripped. 31 requests in 45s from a single browser is entirely client-generated.

### Blast radius

Server-side plan gating is safe: `premiumService.getUserPlan` reads `userSubscriptionsTable` from our own database (`packages/premium/src/subscription.ts`), so sync windows, account/mapping limits and `/api/entitlements` never touch Polar. PR #452's sync windows are not affected by this.

The damage is in the web app, which reads the plan from Polar directly:

- `/dashboard/upgrade` — the loader rethrows anything that is not a 401 (`applications/web/src/routes/(dashboard)/dashboard/upgrade/index.tsx`), so the page a user pays on fails to render outright during a Polar 429.
- `/dashboard/settings` — `loadSubscription` swallows the failure in an empty `catch {}` and returns `undefined`, so a paying Pro user is rendered as free, sees the upgrade CTA, and "Manage plan" routes to `/dashboard/upgrade` instead of the Polar portal.
- `useEntitlements` is unaffected because it prefers the server-side `/api/entitlements` response.

### The endpoint never should have been the plan source

[...]

PR #637, opened 2026-08-13
fix(web): validate the subscription fallback and hold a known billing interval

ridafkih:
Follow-up hardening for #618, which merged before adversarial testing finished. The entitlements fallback body was cast rather than validated, so `{}`, `{plan:"enterprise"}` and `null` all resolved a paying customer to free — the same silent downgrade #619 exists to prevent, one layer down. Adversarial rounds also found the customer-state schema accepted arrays and error envelopes, and that a null recurring interval was coerced to a definite monthly.

- Assert both the fallback and customer-state payloads with an arktype `planSchema`, failing loud in line with the sibling auth-capabilities convention.
- Require `activeSubscriptions` so a 200 error envelope can no longer validate as an empty subscription list.
- Log the upstream Polar failure before falling back, so the motivating outage stops being invisible.
- Hold the last known billing interval as hook state instead of inventing `month`, ending the CTA flip between Manage Subscription and Switch Billing Period during an outage.

Refs #619.

### DIFF
--- a/applications/web/src/hooks/use-subscription.ts
+++ b/applications/web/src/hooks/use-subscription.ts
@@ -99,10 +99,6 @@ export async function fetchSubscriptionStateWithApi(
       throw error;
     }
 
-    console.error(
-      `[subscription] ${CUSTOMER_STATE_PATH} read failed, falling back to ${ENTITLEMENTS_PATH}:`,
-      error,
-    );
 
     const entitlements = entitlementsPlanSchema.assert(
       await fetchApi<unknown>(ENTITLEMENTS_PATH),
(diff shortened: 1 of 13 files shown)


applications/web/src/hooks/use-subscription.ts
```diff
@@ -65,21 +74,39 @@ export function useSubscription(options: UseSubscriptionOptions = {}) {
...
   } catch (error) {
     if (isUnauthorized(error)) {
       throw error;
     }
 
-    const entitlements = await fetchApi<EntitlementsPlanResponse>(ENTITLEMENTS_PATH);
+    console.error(
+      `[subscription] ${CUSTOMER_STATE_PATH} read failed, falling back to ${ENTITLEMENTS_PATH}:`,
+      error,
+    );
```
ridafkih:
```suggestion
```

PR #761, opened 2026-08-14
feat(mcp): annotate every tool with read-only, destructive and open-world hints

ridafkih:


ridafkih:

PR #762, opened 2026-08-14
feat(mcp): add the registry manifest for the hosted MCP server

ridafkih:


server.json
```diff
@@ -0,0 +1,33 @@
...
+      "sizes": ["512x512"],
+      "theme": "light"
+    },
+    {
+      "src": "https://www.keeper.sh/180x180-light-on-dark.png",
+      "mimeType": "image/png",
+      "sizes": ["180x180"],
+      "theme": "dark"
+    }
+  ],
```
ridafkih, line 62: I believe the specification prefers SVG for this, but we should confirm to be sure.

ridafkih:
Will do a review loop on this to make sure we're in-line with the spec.

ridafkih:
Re: SVG icons — the registry schema allows `image/svg+xml`, but its own `src` description warns consumers to "take appropriate precautions when consuming SVGs as they can contain executable JavaScript." In practice that means clients are free to sanitize or refuse them, and a publish is immutable per version, so an icon that renders nowhere is stuck until the next version bump. Keeping the PNGs. If we want scalable later, the schema takes `"sizes": ["any"]` alongside `"mimeType": "image/svg+xml"`.

ridafkih:

PR #766, opened 2026-08-14
feat(web): derive the canonical link from the page path in one head helper

ridafkih:


applications/web/src/routes/(marketing)/index.tsx
```diff
@@ -180,15 +180,12 @@ const FAQ_ITEMS: FaqItem[] = [
...
-    links: [{ rel: "canonical", href: canonicalUrl("/") }],
-    meta: seoMeta({
-      title: "Open-Source Calendar Syncing for Google, Outlook & iCloud",
-      description:
-        "Keep your personal, work, and school calendars in sync automatically. Open source, and works with Google Calendar, Outlook, iCloud, Fastmail and more.",
-      path: "/",
-      brandPosition: "before",
-    }),
+  head: () => seoHead({
+    title: "Open-Source Calendar Syncing for Google, Outlook & iCloud",
```
ridafkih, line 184: We should make sure to change this in the appropriate marketing material updates, because not only does it not actually cover the full providerset, but it leads with the open-source portion which

ridafkih:

PR #768, opened 2026-08-14
feat(web): serve an MCP server card for agent discovery

ridafkih:
SEP-2127 is still an open proposal, so the card carries both the current draft schema fields and the earlier `serverInfo`/`capabilities` pair.


applications/web/src/server/internal-routes.ts
```diff
@@ -47,6 +47,45 @@ const buildProtectedResourceMetadata = (requestOrigin: string) => ({
   scopes_supported: RESOURCE_SCOPES,
 });
 
+// Server Cards let an agent read what the MCP server is and where it listens without
+// connecting first. The proposal behind them (SEP-2127, formerly SEP-1649) is still an
+// open pull request, so the card carries both the identity and transport fields of the
+// current draft schema and the serverInfo/capabilities pair earlier revisions describe.
```
ridafkih: bruh shut up
```suggestion
```

applications/web/src/server/internal-routes.ts
```diff
@@ -47,6 +47,45 @@ const buildProtectedResourceMetadata = (requestOrigin: string) => ({
...
+    {
+      type: "streamable-http",
+      url: `${requestOrigin}/mcp`,
+    },
+  ],
+  serverInfo: {
+    name: MCP_SERVER_NAME,
+    version: MCP_SERVER_VERSION,
+  },
+  // The MCP server registers tools only; it serves neither resources nor prompts.
```
ridafkih: Perhaps an improvement we should make before shipping this.

```suggestion
```

ridafkih:
Re: the capabilities block — resolved by removing it rather than extending it. The current SEP-2127 draft states cards "do not enumerate primitives (tools, resources, prompts) — those remain subject to runtime listing", and `serverInfo` appears in no revision of the proposal at all; identity was always the top-level `name`/`version` the card already carries. So the card no longer claims anything about resources or prompts, and there is nothing to keep in sync if we add them later.

Also moved the card from `/.well-known/mcp/server-card.json` to `/mcp/server-card`, which is the location the draft actually reserves (relative to the Streamable HTTP URL), and renamed it `sh.keeper/keeper` to match the reverse-DNS name `server.json` publishes — the two were describing what looked like two different servers.

ridafkih:

PR #773, opened 2026-08-14
feat(web): move competitor comparisons into a /compare silo

ridafkih:
The old `/blog` comparison paths now permanently redirect to the new `/compare` routes, so existing links and search results keep resolving.


ridafkih:

PR #775, opened 2026-08-14
docs: add a setup guide for Outlook Google Calendar Sync

ridafkih:


ridafkih:
Closing in favour of a rewrite, tracked in KEE-344.

The content here was factually solid — an independent fact-check verified every version number, the £12/yr quota subscription, the fair-usage limits and the Push Sync timings against the OGCS project's own wiki and releases, and the page correctly avoids any successor framing.

The problem was shape, not accuracy. 1,527 body words with a single mention of Keeper.sh at 89% page depth, 16 dofollow links to OGCS-controlled domains against one internal link, and no Keeper.sh entry in its sources block. Because it carried the newest date it also became the first card in the homepage "Latest Guides" block, so the landing page led with a competitor's product name.

The opportunity is real and larger than the rest of the comparison set combined — `ogcs` is 2,900/mo at KD 27, and OneCal, CalendarBridge and SyncThemCalendars each ship a dedicated OGCS page while we ship none. So this returns as `/compare/outlook-google-calendar-sync-alternative` in the silo shape the other comparison pages use: an at-a-glance table first, Keeper.sh named throughout, a switching guide, an in-body CTA, sibling cross-links, and a Keeper.sh line in sources. The honest framing that made this draft credible stays — OGCS is alive and actively maintained, and v3 already drops the classic-Outlook requirement.

PR #786, opened 2026-08-14
fix(web): correct competitor claims and the privacy defaults in docs

ridafkih:
Adversarial review of this stack found four claims that do not survive checking — three about a named competitor, one about our own behaviour.

## Competitor claims

**CalendarBridge added CalDAV on 10 August 2026.** They now cover Fastmail, Nextcloud, Zoho, Yahoo, AOL and self-hosted servers, and those accounts work across the rest of their product. Our comparison page led on CalDAV as the thing they could not do — in the blurb, the connection section, the "where Keeper.sh fits better" list, and the closing recommendation. All four now describe provider coverage as near-parity. The differentiator that replaces it is the one they structurally cannot match: the code is public and the whole thing runs on your own hardware.

The same page said their trust centre was available on request, when it is a public URL, and omitted that SOC 2 Type 2 is in progress.

**OneCal is SOC 2 Type II certified.** The page said its audit was in progress.

**Our yearly option is $42.** The SyncThemCalendars comparison listed it as "not offered"; the other two comparison pages state it correctly.

Both remaining check dates move to 14 August 2026, the date these claims were re-verified against each competitor's own pages.

## Our own privacy defaults

`excludeEventName`, `excludeEventDescription` and `excludeEventLocation` all default to `true`, and `customEventName` defaults to `{{calendar_name}}`. A new calendar copies a block labelled with the source calendar's name and leaves the title, description and location behind.

The new docs and recipes said the opposite:

- `docs/sync-settings.mdx` — "Free copies events with their details as they stand" is the inverse of what ships, and contradicted another line on the same page.
- Same file described **Sync Event Name** as something you turn off, when it ships off.
- `recipes/keep-titles-on-one-side-only.mdx` — step 5 said to leave the work calendar alone, which produces stripped copies in *both* directions.
- `recipes/work-and-personal-both-ways.mdx` — sold Pro as the fix for a problem Free does not have.

Verified against `packages/database/src/database/schema.ts:111-117` and `services/api/src/utils/source-sync-defaults.ts`.


ridafkih:
Closing: every hunk here has been folded into the PR that introduced the line it corrects, so nothing below merges knowingly wrong.

- The Fastmail claim in the founder post and the two how-to intros → #771
- The CalendarBridge, OneCal and SyncThemCalendars claim corrections on the /compare pages, and the related-article ordering → #773 (amended)
- The roundup's CalendarBridge CalDAV/MCP rows and the Kalender Sync / SyncDate verdicts → #774
- The privacy-default corrections across /docs, /guides and /recipes, and the self-hosting copy → #776 (amended)
- The privacy policy, llms.txt and llms-full.txt privacy defaults → #776 (separate commit; those files pre-date the stack)

`git diff` against feat/docs-guides-recipes-triad is empty, and the tip tree is byte-identical to what this branch produced.

PR #787, opened 2026-08-14
fix(web): read alarms separately from their events, and correct the privacy copy

ridafkih:
Adversarial review of this stack found one parsing bug, one structured-data regression, and three claims that don't match the code.

## `VALARM` properties leaked into their events

`parseIcsFile` applied every property line between `BEGIN:VEVENT` and `END:VEVENT` to the event — including lines inside a nested `VALARM`. The component stack was maintained but only consulted on the path where no event was open.

```
SUMMARY:Real event title
DESCRIPTION:Real event description
BEGIN:VALARM
DESCRIPTION:Reminder popup text
END:VALARM
→ description: "Reminder popup text"
```

Google and Outlook both export alarms carrying a `DESCRIPTION`, so this was the common case for files people paste into `/tools/ics-viewer`, not an edge case. Apple-style alarms carry their own `UID` and `SUMMARY`, replacing those too.

## The folding test couldn't fail

It asserted `line.length` — UTF-16 code units — against an ASCII-only fixture. `utf8Length` could have returned a constant and the suite would still have been green. Two new tests measure octets and assert no character is split across a fold.

## Dangling author reference

`BlogPosting.author` and `Organization.founder` became a bare `@id` pointing at a `Person` emitted only on `/about`. Consumers resolve `@id` within a single page's graph, not across pages, so all 21 posts shipped an author with no `@type` and no `name` — strictly less than the inline `Person` it replaced. The reference now carries both.

## Copy that contradicted the code

- **`/privacy`** said copied events "keep their original name, description, and location unless you turn those off". `excludeEventName`, `excludeEventDescription` and `excludeEventLocation` all default to `true` — detail is something you turn **on**. The homepage this stack rewrote already said so, leaving the two pages contradicting each other.
- **Homepage** said attendees are "left behind by default", implying a toggle. There is no attendee field on the outbound payload at all (`packages/calendar/src/core/types.ts`), so they are never copied under any setting.
- **"Every account gets the Pro feature set"** overstated self-hosting, because this stack also added email and passkey sign-in to the plan table and both are gated on commercial mode (`packages/auth/src/index.ts:202,266`).
- The quickstart now names `TRUSTED_ORIGINS`, the variable whose absence fails CSRF checks.
- `PRODUCT_ID` said `Keeper`, which is written into every file a user forwards on.

lint, types, test and unused all pass.


ridafkih:
Closing: every hunk here has been folded into the PR that introduced the line it corrects, so nothing below merges knowingly wrong.

- Homepage copy and FAQ, features page, privacy policy, llms.txt, llms-full.txt, the SoftwareApplication description, the now-dead `brandPosition` option, and the sitemap dates for those pages → #778 (amended)
- Pricing page: "sync mapping" → connection, the $42 annual price, and the honest self-hosting answer → #781
- Self-hosting page copy → #782 (amended); the HTML-cache/sitemap drift and its guard test → #782 (separate commit, extended at #783 and #784 as each adds a page)
- Author `@id` reference carrying `@type` and name, and the about-page descriptions → #783 (amended)
- VALARM parsing, the non-inclusive all-day DTEND, PRODID, and the tool-page copy → #784 (amended)

`git diff` against feat/ics-file-tools is empty, and the tip tree is byte-identical to what this branch produced.

The privacy-policy paragraph keeps both facts it needed: details are stripped by default, and turning them back on is Pro-gated (`PRO_GATED_SOURCE_FIELDS`).

PR #788, opened 2026-08-14
docs(marketing-voice): codify rule 12, lead with the job not the machinery

ridafkih:


ridafkih:

PR #789, opened 2026-08-14
docs(marketing-voice): reference how the category places open source

ridafkih:
Merge after #788 — both change `SKILL.md`, and this PR is deliberately kept to a new file plus one line so the conflict is a single line in the reference list.


ridafkih:

PR #793, opened 2026-08-15
docs: add a capability walkthrough of the MCP calendar toolset

ridafkih:
Adds `applications/web/src/content/blog/what-an-ai-agent-can-do-with-your-calendar.mdx`. "Connect your calendar to an AI assistant" promises everything and specifies nothing — what a reader actually gets is a fixed list of tools. This post is that list, grouped by the job someone would ask for rather than by the shape of the API.

- Walks all 14 tools in four groups — connected calendars and accounts, reading events, writing them, and scheduling and sync — and for each gives the prompt a reader would type, the call it turns into, and the response shape that comes back.
- Tool names, counts, arguments, annotations and the 25-call free daily limit are enumerated from `services/mcp/src/toolset.ts` and `services/api/src/utils/api-rate-limit.ts`, so the list is the toolset rather than an approximation of it.
- Includes what the toolset does not expose — keyword search, attendees, recurrence editing, connecting accounts, other people's availability — and the limits of the consent model. A capability list a reader can trust has to say where it stops.
- Links to the OAuth write-up for how the consent is granted and to `/docs/mcp` for the exact argument names and types, keeping the reference in one place. Adds the post to `applications/web/public/llms.txt`.

Content only — one new MDX post and one line of `llms.txt`, no application code touched.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


ridafkih:

PR #795, opened 2026-08-15
fix(web): serve the OAuth scopes from the constant everyone else uses

ridafkih:
`applications/web/src/server/internal-routes.ts` kept its own copy of the six OAuth resource scopes and served them from `/.well-known/oauth-protected-resource`:

```ts
const RESOURCE_SCOPES = [
  "keeper.read", "keeper.sources.read", "keeper.destinations.read",
  "keeper.mappings.read", "keeper.events.read", "keeper.sync-status.read",
];
```

`packages/auth` and `services/mcp` both read `KEEPER_API_RESOURCE_SCOPES` from `@keeper.sh/constants`. So the discovery document — the thing that tells an MCP client which scopes exist — was the single place free to drift from the scopes the system actually issues and validates.

The two lists agree today. That is precisely why a drift would have gone unnoticed: nothing would fail, the discovery document would just quietly start describing a different server than the one behind it.

`applications/web` now declares `@keeper.sh/constants`, which `packages/calendar`, `packages/auth`, `packages/sync`, `services/api` and `services/cron` already do. The lockfile change is the two lines for that workspace edge.

Found while writing the MCP OAuth post, which had to describe this surface accurately.

### DIFF
--- a/applications/web/package.json
+++ b/applications/web/package.json
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "@better-auth/passkey": "^1.5.5",
+    "@keeper.sh/constants": "workspace:*",
     "@keeper.sh/data-schemas": "workspace:*",
     "@keeper.sh/otelemetry": "workspace:*",
     "@polar-sh/checkout": "^0.2.0",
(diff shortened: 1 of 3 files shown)


ridafkih:

PR #796, opened 2026-08-15
fix(web): call the homepage article block what it actually shows

ridafkih:
The homepage section was headed **"Latest Guides"** and rendered `latestGuides`, which is derived from `articleLibrary` — blog posts and comparison pages. It has never contained anything from the `/guides` silo. The heading promised one thing and the grid showed another.

Renamed to **"Latest content"**, and the code renamed with it so it stops making the same claim:

- `latestGuides` → `latestArticles`, `LATEST_GUIDE_COUNT` → `LATEST_ARTICLE_COUNT`
- `MarketingGuidesSection` / `MarketingGuidesGrid` / `MarketingGuideCard` → `MarketingArticleSection` / `MarketingArticleGrid` / `MarketingArticleCard`
- `marketing-guides-section.tsx` → `marketing-article-section.tsx`

Blog posts and comparisons are the right things to surface there — they are the depth-and-credibility proof, and the `/compare` silo needs the internal links. Only the label was wrong, so the fix is the label rather than the contents.

Also carries the copy sweep's verifier pass, which had been left uncommitted: headings that only parsed if you already knew the product, and three bare "Keeper" mentions.

**This is the refinement PR for this stack** — further adjustments land here rather than being restacked into the PRs underneath.

### DIFF
--- a/applications/web/src/content/compare/best-calendar-sync-tools.mdx
+++ b/applications/web/src/content/compare/best-calendar-sync-tools.mdx
@@ -234,11 +234,11 @@ Connect a work calendar and a personal one. The first block of busy time lands o
 Still weighing tools? These cover the rest of the field.
 
 - [Keeper.sh vs CalendarBridge](/compare/calendarbridge-alternative)
-- [Keeper.sh vs Kalender Sync](/blog/keeper-sh-vs-kalender-sync)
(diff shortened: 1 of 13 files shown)


ridafkih:

PR #798, opened 2026-08-15
fix(web): drop the MCP OAuth build post

ridafkih:
Removes `building-an-mcp-server-with-oauth.mdx`.

The post explained how to build an MCP server with OAuth 2.1 — protected-resource discovery, dynamic client registration, PKCE, audience validation, token lifetimes. That is written for someone building an MCP server, not for someone who needs their calendars to agree with each other. Publishing it codes Keeper.sh as developer tooling, which is the positioning problem the rest of this work has been undoing: every keyword the site ranks for today is an open-source or developer term, and it appears in no buyer-intent query.

Also removes the `llms.txt` entry pointing at it.

## The part worth reviewing

Removing the post removed one copy of a paragraph auditing our own scope enforcement. **The same disclosure survived in `what-an-ai-agent-can-do-with-your-calendar.mdx`**, which said one permission "is what every call is actually checked against" and "the rest are labels".

That paragraph now describes only what a user gets on the consent screen:

> **The permission covers reading and writing together.** Approving Keeper.sh on the consent screen grants create, update, delete and RSVP alongside reading, rather than a read-only half you can accept on its own. Finer-grained permissions are on the roadmap. Until then, a client that asks you to confirm destructive tools individually gives you the same control.

Nothing false, nothing hidden that a user needs in order to decide — and nothing about which internal checks run where.

lint, types, build and 310 web tests pass.


ridafkih:

PR #800, opened 2026-08-15
feat(web): rewrite the landing page around the reader's situation

ridafkih:
Rewrites the landing page so each section leads with the reader's situation rather than the mechanism, and restores the parts of the production page that were working better than the rewrite.

## Copy
- Landing copy culled throughout — plain over clever, no `X, not Y` inversions.
- The three "how it works" steps cut down to one idea each.
- Self-hosting FAQ points at `/self-hosting` instead of restating it, and the guest-list claim is pinned to what the product actually does.

## Pricing
- The feature matrix is back on the landing pricing section, and the plan cards carry bullets again.
- Mobile gets a plan comparison as card bullets rather than a stripped-down table.

## Hero and social proof
- GitHub star count sits next to the hero CTA instead of the header.
- `GithubStarButton` loses its client-side fetch in favour of `github-stars.ts`, so no placeholder ever renders.
- Placeholder testimonials restored behind `testimonials.ts`; the section renders nothing when there is no real testimonial to show.

## MCP
- The MCP feature card names what an agent can actually do with a calendar rather than naming the protocol.

Stacked on #784.


ridafkih:

PR #809, opened 2026-08-15
fix(accounts): scope provider account identity to the owning user

ridafkih:
Follow-up to #805. The production backfill it enabled ran 349 candidates: 316 updated, 32 skipped for reauthentication, and **1 hard failure** on a unique constraint. That failure exposed three defects, each proven here with a test that fails before its fix.

## 1. The unique index was global, not per-user

`calendar_accounts_provider_account_idx` was on `(provider, accountId)` across the whole table. It only ever held because every source row was NULL, and Postgres permits unlimited NULLs in a unique index. Writing real values made it bite — two rows resolved to the same Google account and the second `UPDATE` was rejected.

Two distinct users can legitimately hold one provider account, and could before #805. Migration `0086` scopes the index to `(userId, provider, accountId)`, **and** the source flow now reuses the user's existing row rather than inserting a second. Both were needed; either alone leaves a hole.

## 2. Destination connect silently created nothing — regression from #805

Once a source row carries `accountId`, `saveCalendarDestinationWithDatabase` found it, took the "existing account" branch, found no mapped destination calendar, and returned having created nothing. The user was redirected to `?destination=connected` with no destination. Both paths now route through `upsertAccountAndCalendarWithDatabase`, which already handled create-or-reconnect correctly; the early-return branch was a strictly worse duplicate.

## 3. Migration `0085`'s sibling backfill could only abort

It copied a donor row's `accountId` onto a NULL sibling, which always produces two rows sharing `(provider, accountId)` — violating the index in both its old and new form. Demonstrated against Postgres 17:

```
ERROR:  duplicate key value violates unique constraint "calendar_accounts_provider_account_idx"
DETAIL:  Key (provider, "accountId")=(google, 116651453584579904080) already exists.
```

It matched nothing in production, which is why that deploy survived. The statement is removed and the file carries a comment recording that it ran in production and why it was taken out — a later migration cannot fix this, because drizzle runs all pending migrations in one transaction, so a database that has not yet applied `0085` would roll the whole batch back before any fix-up ran. The intent is now served by `adoptProviderAccountId`, which adopts the id onto an existing row on reconnect.

## Deliberately left undecided

The cross-user ownership check on the destination path is still global. If user A holds a Google account **as a source**, user B is now refused when linking it as a destination — allowed pre-#805, because source rows were invisible to that lookup. Restoring the exact prior meaning needs a reliable "is a destination" predicate, and the obvious one (a row in `source_destination_mappings`) is wrong for a freshly created destination that has no mapping yet. This fails loudly with a clear message rather than corrupting anything, but it is a real behaviour tightening and wants a decision rather than a silent choice.

`0086` does a non-concurrent index swap — sub-second `ACCESS EXCLUSIVE` at this table size, under the migrator's 10s `lock_timeout`.

## Not proven, flagged rather than dropped

`saveCalDAVDestinationWithDatabase` appears to have the identical "no destination created" hole, and `deleteCalendarDestination` deletes the whole `calendar_accounts` row, so removing a destination on a row that also serves as a source may cascade the source calendars away. Both are static reads with no test behind them; #805 did not touch either path.

## Production data

No urgent action. The surviving duplicate pair is Google, which does not need `accountId` for push, so the NULL row still gets channels. Remaining Outlook rows drain on their own as users reconnect.

```sql
SELECT a.id AS holder, b.id AS orphan, a."userId", a.provider, a."accountId", a.email
FROM calendar_accounts a
JOIN calendar_accounts b ON b."userId" = a."userId" AND b.provider = a.provider
                        AND b.id <> a.id AND b."accountId" IS NULL
WHERE a."accountId" IS NOT NULL AND a.provider IN ('google','outlook');
```

https://claude.ai/code/session_01JWminUmH2kMq6fQShPLgvE

### DIFF
--- a/packages/database/drizzle/0085_fixed_kat_farrell.sql
+++ b/packages/database/drizzle/0085_fixed_kat_farrell.sql
@@ -1,14 +1,3 @@
 UPDATE "calendar_accounts"
 SET "accountId" = NULL
 WHERE "accountId" = '';
---> statement-breakpoint
-UPDATE "calendar_accounts" AS target
-SET "accountId" = donor."accountId"
-FROM "calendar_accounts" AS donor
-WHERE target."accountId" IS NULL
-  AND donor."accountId" IS NOT NULL
-  AND donor."id" <> target."id"
-  AND donor."userId" = target."userId"
-  AND donor."provider" = target."provider"
-  AND donor."email" IS NOT NULL
-  AND donor."email" = target."email";
--- a/packages/database/drizzle/0085_fixed_kat_farrell.sql
+++ b/packages/database/drizzle/0085_fixed_kat_farrell.sql
@@ -1,3 +1,9 @@
+-- A second statement shipped here and ran in production, where it matched no rows.
+-- It copied a sibling row's "accountId" onto a row holding NULL, which always produces
+-- two rows sharing ("provider", "accountId") and violates the unique index in both its
+-- pre-0086 and post-0086 form. It could only ever match nothing or abort the migration,
+-- so it was removed rather than left to break databases that had not yet applied 0085.
+-- Adopting a provider account id onto an existing row is handled by adoptProviderAccountId.
 UPDATE "calendar_accounts"
 SET "accountId" = NULL
 WHERE "accountId" = '';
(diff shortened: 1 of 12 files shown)


ridafkih:
Correction to this PR's description.

The second item — "Destination connect silently created nothing" — was described as a production risk. It is not reachable. `/api/destinations/authorize` has no caller anywhere in the repository; the web app's only OAuth entry point is `/api/sources/authorize` (`oauth-preamble.tsx:135`), so `saveCalendarDestinationWithDatabase` and everything beneath it is dead code.

The behaviour was real at the function level, but no user could hit it. That justification was overstated.

The other two findings are unaffected and were genuine: the global `(provider, "accountId")` unique index really did reject the production backfill, and migration `0085`'s sibling statement really could only abort or match nothing.

PR #810, opened 2026-08-15
fix(destinations): stop reconnect creating a duplicate destination calendar

ridafkih:
Regression from #809, already on `main`. Found by an adversarial review of that PR after it merged.

#809 routed both destination paths through `upsertAccountAndCalendarWithDatabase` — correct, since the old early-return branch created nothing at all. But that function treats a calendar as existing only when it already appears in `source_destination_mappings.destinationCalendarId`. A destination that has been connected but not yet wired to a source is invisible to that lookup, so **every reconnect inserts another `<provider> destination` calendar**.

The normal state between connecting a destination and configuring a mapping is exactly that window. Re-running the destination OAuth in it — a scope re-grant, a reconnect, a retried callback — adds a row each time. They stay hidden from `listCalendarDestinations` (which also requires a mapping), so they accumulate silently and then all become live destinations the moment one mapping is created.

## The fix

Match the destination placeholder by its null `externalCalendarId` rather than by mapping membership. Source calendars always set `externalCalendarId` (`oauth-sources.ts:726`, `:899`); the destination placeholder never does. A mapped destination was created by this same insert, so it also has a null value — the new predicate is a superset of the old one and keeps the mapped case working.

## Verification

Reproduced against `main` before fixing, with a pglite test that drives `saveCalendarDestinationWithDatabase` twice:

```
expected 2 to be 1     ← calendars under the account after a second connect
```

It passes with this change. Full suite uncached: **45/45 tasks**, api 716 tests across 87 files.

I also trimmed unused scaffolding from the test as authored (an unused import, client alias and seed helper) so it passes oxlint's `no-unused-vars`.

https://claude.ai/code/session_01JWminUmH2kMq6fQShPLgvE

### DIFF
--- a/services/api/src/utils/destinations.ts
+++ b/services/api/src/utils/destinations.ts
@@ -7,7 +7,7 @@ import {
   syncStatusTable,
 } from "@keeper.sh/database/schema";
 import { REAUTHENTICATION_DESTINATION_GRANT } from "@keeper.sh/constants";
-import { and, desc, eq, inArray, sql } from "drizzle-orm";
+import { and, desc, eq, inArray, isNull, sql } from "drizzle-orm";
 import {
   buildReauthenticationDemandFields,
   getErrorMessage,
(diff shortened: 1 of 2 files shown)


ridafkih:
Correction to this PR's description.

I claimed this was a live regression on `main`. It is not. `/api/destinations/authorize` has no caller anywhere in the repository — the web app's only OAuth entry point is `/api/sources/authorize` (`oauth-preamble.tsx:135`) — so `/api/destinations/callback/:provider` is unreachable, and with it the whole chain this fix sits in: `handleOAuthCallback` → `saveCalendarDestination` → `saveCalendarDestinationWithDatabase` → `upsertAccountAndCalendarWithDatabase`.

The duplicate-insert behaviour is real at the function level and the test proving it is sound, but no request can reach it, so no user was affected. I proved the behaviour without checking reachability.

The change is still correct if that path is ever revived, and harmless meanwhile. A follow-up removes the dead chain entirely.

PR #842, opened 2026-08-17
feat(two-way): let the user grant writes to the meetings they organise

ridafkih:


SSD2019:
Thank you so much for creating this program. I'm eagerly waiting for the two-way sync. Is the branch ready to use selfhosted?

Issue #853, opened 2026-08-18
Pin or revalidate DNS records before making CalDAV/ICS fetch


kdalal-vulncheck:
I’m a vulnerability analyst at [VulnCheck](https://www.vulncheck.com/), an exploit intelligence company and research CVE Numbering Authority (CNA), where I'm one of several folks who manage our coordinated vulnerability disclosure (CVD) program.

A security researcher recently [reported](https://www.vulncheck.com/advisories/report) a potential vulnerability impacting ridafkih/keeper.sh and VulnCheck is coordinating this disclosure. 

The following CVE ID have been provisionally reserved for reference purposes. Reservation does not constitute publication; the IDs will remain in a non-public RESERVED state with no technical details visible in the CVE database until we have aligned on CVE-eligibility, disclosure details, and a timeline, or until the 120-day coordinated disclosure window closes.
 
CVE-2026-75583 - DNS-Rebinding Bypass of SSRF Guard (BLOCK_PRIVATE_RESOLUTION) in keeper.sh Calendar-Fetch Layer

VulnCheck follows a [120-day disclosure policy](https://www.vulncheck.com/vulnerability-disclosure-policy), meaning we afford vendors/maintainers up to 120 days from the time of receiving the report to address the issues before publication of CVE records and third-party advisories. For this vulnerability, the 120-day deadline falls on December 16, 2026.

___________________________________________________________________________

DNS-Rebinding Bypass of SSRF Guard (BLOCK_PRIVATE_RESOLUTION) in keeper.sh Calendar-Fetch Layer

Package: ridafkih/keeper.sh Tested Commit: 58384b136477f8f82d724c52417020dac8f2acf1 Affected File: packages/calendar/src/utils/safe-fetch.ts (shared by services/api and services/cron) Reporter: Lazizbek Djurayev (Haad TC) CWE: CWE-367: Time-of-check Time-of-use Race Condition → CWE-918: Server-Side Request Forgery CVSS 3.1: AV:N/AC:L/PR:L/UI:N/S:C/C:L/I:N/A:N -- 5.0 Medium (S:C: SSRF crosses from application to internal network; C:L: reachability confirmed, content exfiltration not independently traced)

Affected Files
packages/calendar/src/utils/safe-fetch.ts
  validateHostResolution() lines 77-92:
    const addresses = await resolveAllAddresses(hostname);   // CHECK: independent DNS lookup
    for (const address of addresses) {
        assertUnicastAddress(address, "...");
    }
[...]

PR #857, opened 2026-08-18
fix(calendar): rebind DNS resolution when guard is active

ridafkih:
Pins each request to the address the SSRF guard actually validated, closing the DNS-rebinding bypass in #853


packages/calendar/src/utils/safe-fetch.ts
```diff
@@ -74,32 +75,79 @@ const validateProtocol = (protocol: string): void => {
...
+  host: parsed.host,
+  serverName: pinnedServerName(parsed),
+  urls: addresses.map((address) => pinnedUrl(parsed, address)),
+});
+
+/*
+ * The validated address is carried into the connection because Bun resolves the
+ * hostname again at connect time; a rebinding resolver would otherwise answer a
+ * public address to the check and a private one to the socket (CVE-2026-75583).
+ */
```
ridafkih:
```suggestion
```

packages/calendar/src/providers/caldav/shared/client.ts
```diff
@@ -162,6 +162,36 @@ const getDisplayName = (name: unknown): string => {
   return "Unnamed Calendar";
 };
 
+/*
+ * A PROPFIND response can name any origin in a <d:href>, and tsdav resolves that href
+ * into the collection url it then requests with the account credentials. Binding every
+ * url the server hands back to the origin of the account's serverUrl keeps those
+ * credentials, and the stored calendar url, on the server the user named.
+ */
```
ridafkih:
```suggestion
```

PR #911, opened 2026-08-20
fix(auth): generate a username for social sign-ups

NatnaelTaddese:
## What

Adds `mapProfileToUser` to the Google and Microsoft social providers so a new user created via social sign-in gets a generated `username`.

## Why

In self-hosted mode (`COMMERCIAL_MODE` unset) the `username-only` plugin adds a **required, unique** `username` column to the user model (`packages/auth/src/plugins/username-only/utils/schema.ts`). Google/Microsoft profiles don't supply a username, so first-time social sign-in failed during user creation — the OAuth callback threw `APIError: username is required` and redirected the browser to `/api/auth/error?error=username_is_required`.

## How

- New `generateSocialUsername(profile)` helper derives a username from the sanitized email local-part (falling back to the profile name, then `"user"`), constrained to the plugin's allowed charset `[a-z0-9._-]`, plus a short random suffix (`randomUUID`) for uniqueness against the unique constraint.
- Wired into both `socialProviders.google` and `socialProviders.microsoft`. Microsoft falls back to `preferred_username` when `email` is absent.

Example outputs: `dotphicc-9221f333`, `first.last-tag-3a9d3ec4`, and `user-5c594013` for empty/non-Latin profiles.

## Notes for reviewers

- Only affects **new** social users; `mapProfileToUser` runs at account creation, so returning users and account-linking are unchanged.
- Username/password (`username-only`) sign-up is unaffected — it still validates its own input.
- `bun run types` passes for `@keeper.sh/auth`.

## Testing

Self-hosted dev with `GOOGLE_CLIENT_ID`/`GOOGLE_CLIENT_SECRET` set: "Continue with Google" for a brand-new account now creates the user and lands on `/dashboard` instead of redirecting to the `username_is_required` error page.

### DIFF
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { type } from "arktype";
 import { betterAuth } from "better-auth";
 import { createAuthMiddleware } from "better-auth/api";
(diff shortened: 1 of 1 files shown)


NatnaelTaddese:
@ridafkih  before this goes further, I think this is really a design question, and I'd rather get your call on the approach.

Context I found:
- `username` is required+unique only in self-hosted mode (`usernameOnly()` is pushed behind `if (!commercialMode)`), so username/password is clearly the intended baseline credential there.
- But the social buttons aren't gated by mode — `resolveAuthCapabilities` gates email/passkeys/password-reset on `commercialMode`, yet leaves `socialProviders.google/microsoft` on whenever creds are set, and `getEnabledSocialProviders` doesn't check `credentialMode`. So the self-hosted login page intentionally shows "Continue with Google", but a brand-new social user 500s at the callback with `username_is_required`.

So: is Google/Microsoft sign-in meant to create a self-hosted account, or is it intended for linking to an existing username/password account only?

ridafkih:
@NatnaelTaddese I'll leave this one open for now, since typically self-hosted users are running credential-based login anyways, and I need to overhaul and add OIDC/OAuth so people can use their Authentik/Okta credentials.

PR #912, opened 2026-08-20
revert: webhook ingest preempting a fleet pass on the same calendar

ridafkih:


ridafkih:
Superseded — folding this into the lane-disjoint fix rather than reverting separately. Branch kept as a fallback.

PR #916, opened 2026-08-20
perf(cron): read the calendar row once per ingest

ridafkih:


ridafkih:
Superseded — rebuilding against the current tree rather than merging a pre-#913 restructuring of runSourceIngest.

PR #917, opened 2026-08-20
perf(cron): cache the stored event count for the weight estimate

ridafkih:


packages/database/src/database/schema.ts
```diff
@@ -140,6 +140,14 @@ const calendarsTable = pgTable(
     name: text().notNull(),
     originalName: text(),
     syncToken: text(),
+    /*
+     * Cached count of event_states rows for this calendar, maintained by the
+     * ingest flush transaction so the pre-fetch weight estimate needs no
+     * count(*). Null means "not yet cached"; readers fall back to counting.
+     * May over-count (inserts use ON CONFLICT DO UPDATE), which is harmless:
+     * the weight estimate it feeds is clamped and deliberately conservative.
+     */
```
ridafkih, line 149:
```suggestion
```

actually stfu

PR #927, opened 2026-08-21
fix(outlook): propagate recurrence and verify destination series by id

lucky-lusa:
## Summary

`serializeOutlookEvent` never included a `recurrence` field, so recurring source events (e.g. a biweekly Monday evening series) were pushed to Outlook destinations as a single one-time event pinned to the series' original start time — often already in the past and never visible again. Google and CalDAV destinations already propagate `recurrenceRule`; Outlook did not.

Adds `buildOutlookRecurrence` to translate the internal RRULE-like `recurrenceRule` into Microsoft Graph's pattern/range recurrence shape, and extends `outlookEventSchema` with the `recurrence` field. Sub-daily frequencies (SECONDLY/MINUTELY/HOURLY), which Graph has no equivalent for, fall back to a single non-recurring push instead of being dropped.

With recurrence now attached, a second bug surfaced immediately in production: `listRemoteEvents` filtered by a lookback window, but a recurring series' own start is always its first (possibly long-past) occurrence, never any of its future ones. Any pushed recurring series whose original start predates the lookback window was invisible to this existence check on every subsequent sync, got marked stale, and was deleted and recreated from scratch every single cycle.

The fix is a `CalendarSyncProvider.verifyEventsExist(deleteIds)` capability: a direct by-id GET, independent of any time window, category inheritance on recurring instances, or pagination. `listRemoteEvents` stays a bounded, best-effort discovery pass used only to find orphaned keeper events with no mapping — a reasonable place to accept a rare miss. But it is no longer the source of truth for whether an *already-mapped* event still exists: before treating a mapping as gone, the destination is asked directly by id for anything the discovery pass didn't surface.

## Test plan
- [ ] Existing tests pass
- [ ] Push a biweekly recurring event to an Outlook destination: recurrence is visible and correct in Outlook
- [ ] Push a recurring series whose original start is well outside the lookback window, then re-sync: the mapping survives instead of being deleted and recreated

### DIFF
--- a/packages/calendar/src/core/sync-engine/types.ts
+++ b/packages/calendar/src/core/sync-engine/types.ts
@@ -22,13 +22,6 @@ interface CalendarSyncProvider {
   getRemoteEventsByIds?: (eventIds: string[]) => Promise<RemoteEvent[]>;
   getThrottleMetrics?: () => ProviderThrottleMetrics;
   getSyncDiagnostics?: () => Record<string, number | string>;
-  /**
-   * Direct by-id existence check for destination events that a bulk listRemoteEvents pass
-   * didn't surface. Optional: only providers whose listing endpoint can miss events it should
-   * find (e.g. a bounded window that could disagree with the destination API's own expansion of
-   * a recurring series) need this as a fallback before treating a mapping as genuinely gone.
-   * Returns only the events that were confirmed to still exist.
-   */
   verifyEventsExist?: (deleteIds: string[]) => Promise<RemoteEvent[]>;
 }
 
--- a/packages/calendar/src/core/sync-engine/types.ts
+++ b/packages/calendar/src/core/sync-engine/types.ts
@@ -22,6 +22,14 @@ interface CalendarSyncProvider {
   getRemoteEventsByIds?: (eventIds: string[]) => Promise<RemoteEvent[]>;
   getThrottleMetrics?: () => ProviderThrottleMetrics;
   getSyncDiagnostics?: () => Record<string, number | string>;
+  /**
+   * Direct by-id existence check for destination events that a bulk listRemoteEvents pass
+   * didn't surface. Optional: only providers whose listing endpoint can miss events it should
+   * find (e.g. a bounded window that could disagree with the destination API's own expansion of
+   * a recurring series) need this as a fallback before treating a mapping as genuinely gone.
+   * Returns only the events that were confirmed to still exist.
+   */
+  verifyEventsExist?: (deleteIds: string[]) => Promise<RemoteEvent[]>;
 }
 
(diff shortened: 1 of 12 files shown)


packages/calendar/src/core/sync-engine/types.ts
```diff
@@ -22,6 +22,14 @@ interface CalendarSyncProvider {
   getRemoteEventsByIds?: (eventIds: string[]) => Promise<RemoteEvent[]>;
   getThrottleMetrics?: () => ProviderThrottleMetrics;
   getSyncDiagnostics?: () => Record<string, number | string>;
+  /**
+   * Direct by-id existence check for destination events that a bulk listRemoteEvents pass
+   * didn't surface. Optional: only providers whose listing endpoint can miss events it should
+   * find (e.g. a bounded window that could disagree with the destination API's own expansion of
+   * a recurring series) need this as a fallback before treating a mapping as genuinely gone.
+   * Returns only the events that were confirmed to still exist.
+   */
```
ridafkih:
```suggestion
```

packages/calendar/src/providers/outlook/destination/provider.ts
```diff
@@ -326,6 +326,14 @@ const createOutlookSyncProvider = (config: OutlookSyncProviderConfig) => {
...
   };
 
+  /**
+   * Best-effort discovery of keeper-tagged events not tracked by any mapping, so they can be
+   * cleaned up as orphans. This is intentionally bounded to a time window: a recurring series'
+   * own start/dateTime is its first (possibly long-past) occurrence, so a series that started
+   * before the window can be missed here. That's an acceptable gap for orphan discovery (a
+   * stray duplicate lingers a bit longer) - it must NOT be relied on to confirm whether an
+   * *existing* mapping's event still exists, which is what verifyEventsExist is for.
+   */
```
ridafkih:
```suggestion
```

packages/calendar/src/providers/outlook/destination/provider.ts
```diff
@@ -403,13 +411,62 @@ const createOutlookSyncProvider = (config: OutlookSyncProviderConfig) => {
 
   const getThrottleMetrics = (): ProviderThrottleMetrics => ({ ...throttleMetrics });
 
+  /**
+   * Authoritative existence check for events we already track a mapping for for: a direct
+   * GET by the destination's own event id, independent of any time window, category
+   * inheritance on recurring instances, or pagination. This is what determines whether an
+   * already-pushed event should be treated as still existing - listRemoteEvents' bounded
+   * window is not reliable enough for that call for a long-running recurring series.
+   */
```
ridafkih:
```suggestion
```

PR #934, opened 2026-08-21
feat(sync): suppress mirroring events the destination already holds natively

ridafkih:
When two connected calendars are invited to the same event, each calendar
holds a native copy under the same iCal UID, and the destination sync
previously mirrored each copy onto the other calendar anyway, duplicating
the event.

The destination sync now builds an occurrence index of the destination
calendar's own ingested event states — expanding native recurring series
over the local-read window — and drops any materialized source occurrence
whose UID and instants (at second granularity) already exist natively.
Existing duplicate mirrors are cleaned up by the normal orphan-mapping
reconciliation path. Matching is per-occurrence, never rule-level, so
declined instances, moved overrides, and stale series on one side keep
their slots covered. Every fallback (over-budget series, duplicate
masters, missing native data, keeper-authored rows) degrades to the
previous mirroring behavior rather than wrongly suppressing.

Work in progress: adversarial review round 3 pending.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017CKw53yFqDnwQyNjhCmPSx


packages/sync/src/sync-user.ts
```diff
@@ -325,6 +332,92 @@ const getBoundingSourceAuthorityWindow = (
...
+ */
+const suppressNativeDuplicates = async (
+  context: NativeDuplicateSuppressionContext,
+): Promise<NativeDuplicateFilterResult> => {
+  const suppressionWindow = resolveNativeSuppressionWindow(
+    context.destination,
+    context.localReadWindow,
+  );
+  if (!suppressionWindow) {
+    return { events: context.events, suppressedCount: 0 };
```
ridafkih, line 411: Monkey-patching different filtering logic like this is not going to scale, we are going to have to come up with an actual, decent, API interface for this or else it'll keep ballooning into what is currently already a clusterfuck of disgusting branching and logic chains. c:

packages/calendar/src/core/events/native-duplicates.ts
```diff
@@ -0,0 +1,442 @@
...
+  try {
+    return parseStoredRecurrenceForMaterialization({
+      eventId: row.id,
+      exceptionDates: row.exceptionDates,
+      recurrenceId: row.recurrenceId,
+      recurrenceRule: row.recurrenceRule,
+    });
+  } catch {
+    return null;
+  }
```
ridafkih, line 122: We get zero observability into this if it fails, now.

PR #938, opened 2026-08-23
feat(web): show events in the dashboard calendar

NatnaelTaddese:
## Summary

Events now render in the dashboard's calendar pane, in the same card style as the `/dashboard/events` page, with stacking and height-responsive content modelled on qali's week view.

- **Shared event card** — `EventCard` moves out of the events route into `features/dashboard/components/event-card.tsx` with a `list` layout (events page, unchanged) and a `grid` layout for the calendar.
- **Week grid** — each day is a memoised `DayColumn` that lays out its events by day fraction. Overlapping events cascade (indent right, later start paints on top) and tile into equal columns when two start within 45 min (`event-layout.ts`, ported from qali). Positions come from the local wall clock, so DST days keep noon on the noon rule.
- **Height-adaptive cards** — the grid card is a CSS size container; `event-compact/short/roomy/tall` custom variants step the content down with the card's height (title only at 15–30 min, time + title at 1 h, description from 75 min), plus `event-narrow` for tiles too thin for a time range.
- **Data** — `useEventsInRange` with a week-quantised window (`getWeekFetchRange`), so scrolling inside it never refetches and a page in either direction stays covered by the previous window; `bucketEventsByDay` groups the window's events by local day.
- **All-day events** — `feat(api)` exposes `isAllDay` on the event response; the web keeps all-day events out of the time grid and shows them as pills in a band under the day row (0–2 rows, "+N more").
- **Month view** — day cells list pills (all-day first) in as many rows as fit, folding the rest into "+N more".
- **Scroll sync** — column rules are painted on the day cells and the header's day row follows the grid through a CSS scroll timeline (with the `scrollLeft` mirror as a fallback), so cards, rules and day labels move as one during a swipe.
- **Opaque fills** — the dark-mode event background mixes against the page colour instead of transparent; past cards dim their content rather than the element, on a muted fill, so stacks stay legible.

## Verification

- Per commit: `bun run lint`, `bun run types`, `bun run test` (web: 466 tests, incl. layout, bucketing, fetch-window and time-format suites; api: 702) and `bun run unused` (knip) at the root.
- Live on `/dashboard`: cards sit on their hour rules; bands resolve on real durations; overlaps cascade/tile; all-day events appear only in the band; one `/api/events` request per window, none while scrolling inside it; month switch fetches exactly the 42 grid days; header row translate equals `-scrollLeft` to the sub-pixel at multiple offsets.

## Before/After

| Before | After |
|:---:|:---:|
| events as plain rows | events as cards |
image

## Notes

- Scroll-timeline fallback (engines without `animation-timeline`) is the previous `scrollLeft` mirror and wasn't exercised here.
- Column rules no longer run through a vertical overscroll bounce (they end at the content edge, under the existing edge fade).

### DIFF
--- a/applications/web/src/components/ui/primitives/text.tsx
+++ b/applications/web/src/components/ui/primitives/text.tsx
@@ -17,6 +17,8 @@ const text = tv({
       inverseMuted: "text-foreground-inverse-muted",
       default: "text-foreground",
       danger: "text-red-500",
+      event: "text-event",
+      eventMuted: "text-event-muted",
     },
     align: {
       center: "text-center",
@@ -34,7 +36,7 @@ const text = tv({
 type TextProps = PropsWithChildren<{
   as?: "p" | "span";
   size?: "base" | "sm" | "xs";
-  tone?: "muted" | "disabled" | "inverse" | "inverseMuted" | "default" | "danger" | "highlight";
+  tone?: "muted" | "disabled" | "inverse" | "inverseMuted" | "default" | "danger" | "highlight" | "event" | "eventMuted";
   align?: "center" | "left" | "right";
   className?: string;
   style?: CSSProperties;
--- a/applications/web/src/components/ui/primitives/text.tsx
+++ b/applications/web/src/components/ui/primitives/text.tsx
@@ -17,8 +17,6 @@ const text = tv({
       inverseMuted: "text-foreground-inverse-muted",
       default: "text-foreground",
       danger: "text-red-500",
-      event: "text-event",
-      eventMuted: "text-event-muted",
     },
     align: {
       center: "text-center",
@@ -36,7 +34,7 @@ const text = tv({
 type TextProps = PropsWithChildren<{
   as?: "p" | "span";
   size?: "base" | "sm" | "xs";
-  tone?: "muted" | "disabled" | "inverse" | "inverseMuted" | "default" | "danger" | "highlight" | "event" | "eventMuted";
+  tone?: "muted" | "disabled" | "inverse" | "inverseMuted" | "default" | "danger" | "highlight";
   align?: "center" | "left" | "right";
   className?: string;
   style?: CSSProperties;
--- a/applications/web/src/index.css
+++ b/applications/web/src/index.css
@@ -19,6 +19,12 @@
 @custom-variant event-short (@container event-card (height < 43px));
 @custom-variant event-roomy (@container event-card (59px <= height < 75px));
 @custom-variant event-tall (@container event-card (height >= 75px));
+/* Narrow cards — overlapping events tiled three across in a week column —
+   can't show a time range's both ends, so the time line goes and the title
+   tightens instead. Declared last on purpose: custom variants emit in
+   declaration order, so `event-narrow:hidden` outranks the roomy/tall
+   description clamps on the same element. */
+@custom-variant event-narrow (@container event-card (width < 72px));
 
 html {
   @apply w-full min-h-full bg-background overflow-x-hidden;
--- a/applications/web/src/index.css
+++ b/applications/web/src/index.css
@@ -2,6 +2,21 @@
 
 @custom-variant pointer-hover (@media (hover: hover));
 
+/* Calendar event cards are sized by their event's duration, so their content
+   steps down as the card gets shorter. A grid card is a size-query container
+   named `event-card` (see the grid layout in `EventCard`); these variants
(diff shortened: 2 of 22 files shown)


applications/web/src/components/ui/primitives/text.tsx
```diff
@@ -17,6 +17,8 @@ const text = tv({
       inverseMuted: "text-foreground-inverse-muted",
       default: "text-foreground",
       danger: "text-red-500",
+      event: "text-event",
+      eventMuted: "text-event-muted",
```
ridafkih: One-off variants like this justify their own components, otherwise the `Text` components (or other atomic primitives) will end up getting these huge lists of variants. 

These components are meant to be very generic, so they shouldn't include variants like this.

applications/web/src/index.css
```diff
@@ -2,6 +2,30 @@
...
+   select on its rendered height. The thresholds assume the week grid's 48px
+   hour, Tailwind's 16px text-xs and 20px text-sm lines and the grid card's
+   4px vertical padding — the band table in `event-card.tsx` has the fit
+   arithmetic. Each floor sits 1px under its band's content height: a card's
+   height is a percentage of the day, which can land a hair under the exact
+   pixel and would otherwise drop the card into the band below; the 1px it
+   may clip is padding, never glyphs. The container is named deliberately:
+   height queries need a `size` container, and an unnamed query binds to
+   whichever ancestor happens to be one, so a future size container around
+   the card could silently capture these. */
```
ridafkih: I like to do one pass at the end of my AI coding sessions to get rid of all these comments, Claude loves to write novels above simple snippets.

applications/web/src/index.css
```diff
@@ -87,6 +111,17 @@ html {
   --ui-ring: var(--color-blue-500);
   --ui-template: var(--color-purple-600);
   --ui-template-muted: color-mix(in srgb, var(--ui-template), transparent 50%);
+  --ui-event: var(--color-blue-900);
+  --ui-event-muted: color-mix(in srgb, var(--ui-event), transparent 30%);
+  --ui-event-background: var(--color-blue-100);
```
ridafkih: Same thing sort of applies for this, I personally like to really tightly couple the implementation with the styles, and have the _component_ be the single soruce of truth for all styles and markdown for a component. Big reason I like Tailwind.

applications/web/src/index.css
```diff
@@ -244,6 +284,57 @@ html {
...
+   width less the viewport's, set inline by `WeekGrid`) in step with it, on
+   the compositor. `WeekGrid` keeps the `scrollLeft` mirror as the fallback
+   for engines without scroll timelines. */
+.calendar-strip-scope {
+  timeline-scope: --calendar-strip;
+}
+
+.calendar-strip-scroller {
+  scroll-timeline: --calendar-strip x;
+}
```
ridafkih: Ditto, Tailwind has been defeated if we're just going to make global class names to apply styles to

applications/web/src/index.css
NatnaelTaddese: yeah it's annoying. I'll try to remove on the way too

applications/web/src/index.css
NatnaelTaddese: I was following the `--ui-destructive-*` / `--ui-template-*` pattern that's already in this file, these are theme values more than component styles, the dark swap happens in the one media query. The card still owns all its styling through the `bg-event-*` / `text-event` utilities, the tokens are just the palette underneath (same idea as `--color-background`). Happy to inline them with dark: variants if you'd still rather, it'd just theme the event colors differently from the other families.

applications/web/src/index.css
NatnaelTaddese: fair, it was since I couldn't inline the @keyframes using tailwind. but using motion for this problem seems to be the better approach

applications/web/src/index.css
ridafkih: @NatnaelTaddese Could we name them generically, instead? Rather than component-based semantics, `--ui-callout-blue` would be better, because it prevents us from "crossing boundaries" later.

NatnaelTaddese: yeah agreed, will fix

NatnaelTaddese: I have moved event colors, considering we'd have more variations, into a component preset config, using scoped variables.

ibehren1:
I love the look of this and having this view available in the web app.+1

PR #940, opened 2026-08-23
perf(caldav): raise Fastmail's request budget on its own throttle telemetry

ridafkih:
## The evidence #932 was built to produce

#932 split the per-account CalDAV budget per provider and deliberately held both at **240** — a number borrowed from Microsoft's four-requests-per-second-per-mailbox documentation, never researched for Fastmail — with a test asserting they stay there *"until telemetry says otherwise."*

Telemetry now says otherwise. From 90 minutes of production ingests:

| signal | value |
|---|---|
| our limiter's refusals, saturated ingest | ~3,400 median |
| Fastmail's own throttles, same window | 181 |
| `window_occupancy_max` on the pinned account | **240** — at the cap |
| `ratelimit.queue_depth_max` | 643 |
| `provider.request_count` on those ingests | 1–2 |
| `retry_after_honoured_count` | **0** of 181 |

On *successful* Fastmail ingests, our own limiter is about **5%** of the work while the provider's retry backoff is **23%**. The constraint that shows up in the median is Fastmail's, not ours — but the pathological tail is entirely self-inflicted: an account pinned at its own ceiling with a 643-deep queue, making one request.

Fastmail publishes no CalDAV request rate. The ceiling was never theirs.

## Why 480 and not higher

Raised as a **bounded step, not removed**. Fastmail does throttle, and it sends **no `Retry-After`** — every one of those 181 throttles fell through to our local floor — so we are still guessing the backoff. Doubling is reversible and observable; removing the ceiling would not be.

`ratelimit.provider_throttle_count` and the honoured-versus-floor split are the signal to watch. If provider throttling climbs after this lands, step it back.

**iCloud stays at 240**, where Apple demonstrably throttles — that's why the 503 handling exists.

## Verification

`bunx turbo run test lint types --force --continue` → **45/45**. The budget test now asserts iCloud's 240 and Fastmail's raised value separately, with the reason in each name.


ridafkih:
Closing — the evidence I opened this on does not support it, and an adversarial review showed the change has negative expected value.

**The headline number was a unit error.** I wrote that our limiter refused ~3,400 times per saturated ingest against Fastmail's 181 throttles. Those counters are not commensurable:

- `ratelimit.throttled_count` is incremented **inside** the acquire retry loop, so it fires once per *poll iteration*
- `ratelimit.provider_throttle_count` fires once per *429/503 response*

3,400 ticks across 643 waiters is ~5.3 polls each — consistent with the limiter being only **5% of accounted time on the median ingest**, which the same telemetry showed and which I argued past. The commensurable denominator, `ratelimit.acquire_count`, is emitted one line above the tick. This repo's own suite already documents the mismatch: `rate-limiter-fast-path-telemetry.test.ts` asserts `acquire_count === 1` alongside `throttled_count === 2` for a single acquire.

**The queue is not costing throughput.** Running the actual `ACQUIRE_SCRIPT` Lua against a sorted-set emulator driven by the real client loop gives steady-state grants of 240/min at limit 240 and 480/min at limit 480, at every concurrency tested. The FIFO handoff, head grace and poll ceiling impose no ceiling of their own — achieved rate tracks the constant. So this does not drain a starved queue; it doubles the request rate reaching a provider that is already pushing back.

**The trade prices out negative.** Eliminating *all* limiter wait buys ~5% of ingest wall-clock. Provider retry backoff is already 23%, and with no `Retry-After` every throttle costs our invented floor of 2s/4s/8s — 181 throttles is at minimum 362s. Roughly 40 additional provider throttles, a 22% increase, erases the entire gain. Doubling the offered rate against a throttling provider will plausibly do considerably more than 22%. Provider retries also re-charge the limiter, so throttles consume the budget as well.

The saturated account remains a real problem — 643 waiters against 1–2 actual requests is a fairness or accounting issue inside one account's key, not evidence the ceiling is too low. That is worth fixing on its own terms.

`ratelimit.provider_throttle_count` and the honoured-vs-floor split from #932 remain the right signals; they simply do not say what I claimed they said.

PR #952, opened 2026-08-26
fix(auth): tear down provider state before deleting a user

ridafkih:


ridafkih:
Superseded by #958 (incident fix) and #961 (connect path).

Issue #875, opened 2026-08-19
Icloud shared calendar notication flood


xairam:
Hello
 I just tested it and connected an caldav calenda and an shared icloud calendar(I created it and shared it with my wife).

the caldav calender sends events to the icloud calendar and thas working fine. 
But my Wife getting notifcations that an event got deleted/added on a minutes base, wich is annoying. 

Is this a bug or can I se the update rate?

Best Regards, perfect software so far.

ridafkih:
@xairam Were the notifications sent once (with the big flood of new events being added to the calendar) or are they being delivered on a consistent basis?

ridafkih:
It would be useful to know the number of notifications being delivered to your wife’s phone. :) Once I have that, I can begin debugging.

xairam:
Its only 1 event in the test calendar. 

looks like many deletes and the last was an add notification.
So far I can see its every 10 Min. 2 Notifications. 


I am using the selfhosted version.

ridafkih:
@xairam What's your CalDAV server, and this is self-hosted, right?

xairam:
Yes also selfhosted. 

I am using the Grommunio Appliance.
https://github.com/grommunio/grommunio-dav

ridafkih:
@xairam Could you give v2.21.0 a go? See if that fixes it. 
If not, it's likely the IDs are changed under the source ingestion: will address that when it's confirmed. c:

xairam:
Unfortunately not it  keeps going. 


<img width="946" height="2048" alt="Image" src="https://github.com/user-attachments/assets/0ad670f8-0ccc-4d47-9574-15178ec5b9a9" />

ridafkih:
@xairam Omg, tell her I'm so sorry - I'll spin up the server and check against it myself, it might be mangling IDs or something.

xairam:
@ridafkih if you need i can give you an account to test with.

ridafkih:
> @ridafkih if you need i can give you an account to test with. 

Thank you, but no need! I have reproduced the issue by spinning up my own instance. It looks like there's some normalization that happens that triggers a re-sync.

I'll fix it after work today. :)

ridafkih:
@xairam Going to require a little more work, I was camping over the weekend: but back today! I can have a look, and should have a fix sometime this week.

xairam:
@ridafkih Do you have a version i can test this with?

PR #955, opened 2026-08-26
fix(caldav): never resolve a create conflict by deleting a stranger's event

ridafkih:
Our CalDAV object names are derived from the source event, so two customers mirroring the same event into one shared calendar land on the same href. When the second create conflicts, `recoverCreateConflict` fetched whatever was already standing at that href and — unless it was byte-for-byte the mirror we were about to write — **deleted it with its own ETag** and put ours in its place, then returned `conflictResolved: true, success: true`.

On a shared calendar that destroys the other customer's event, and each side does it to the other on every pass. The push reports success either way.

## What changed

The object already there now has to carry this event's uid before anything is removed. Anything else fails the push and leaves both copies alone.

Failing costs us a mirror we could not write, which the next cycle reports and an operator can see. The alternative cost somebody their event, silently, and told us it had worked.

## Why it went unnoticed

This path had **no test at all** — the three files mentioning `CalDAVCreateConflictError` only construct the error, never the recovery. It is unchanged from `main` and independent of the #875 churn work, which is why fourteen rounds of review on that branch never reached it: every round attacked the code that branch changed.

## Verification

Three cases, red-first against `main`:

- an object carrying a uid that is not this event's — **was deleted**, now is not
- an object we could not parse at all — left alone
- our own object whose content has drifted — still replaced, `conflictResolved: true`

Mutation-controlled: removing the guard turns the first case red again.

Workspace gate: `bunx turbo run test lint types --force` → **45/45**.

### DIFF
--- a/packages/calendar/src/providers/caldav/destination/provider.ts
+++ b/packages/calendar/src/providers/caldav/destination/provider.ts
@@ -120,6 +120,19 @@ const recoverCreateConflict = async (
     return;
   }
 
+  /*
+   * Our object names are derived from the source event, so two customers mirroring the same event
+   * into one shared calendar land on the same href. The object already standing there is only ours
+   * to replace if it says so: anything else is somebody else's event, and deleting it to make room
+   * destroys a copy we were never asked to manage. Failing here costs us a mirror we could not
+   * write; the alternative costs them the event itself.
+   */
+  if (remoteEvent?.uid !== uid) {
+    throw new Error(
+      `CalDAV object ${uid}.ics carries uid ${remoteEvent?.uid ?? "none"}, so it is not this event's mirror`,
+    );
+  }
+
   if (!existing.etag) {
     throw new Error(`CalDAV event ${uid} already exists but has no ETag for a safe recreation`);
   }
--- a/packages/calendar/src/providers/caldav/destination/provider.ts
+++ b/packages/calendar/src/providers/caldav/destination/provider.ts
@@ -120,13 +120,6 @@ const recoverCreateConflict = async (
     return;
   }
 
-  /*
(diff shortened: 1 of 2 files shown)


packages/calendar/src/providers/caldav/destination/provider.ts
```diff
@@ -120,6 +120,19 @@ const recoverCreateConflict = async (
     return;
   }
 
+  /*
+   * Our object names are derived from the source event, so two customers mirroring the same event
+   * into one shared calendar land on the same href. The object already standing there is only ours
+   * to replace if it says so: anything else is somebody else's event, and deleting it to make room
+   * destroys a copy we were never asked to manage. Failing here costs us a mirror we could not
+   * write; the alternative costs them the event itself.
+   */
```
ridafkih:
```suggestion
```

packages/calendar/tests/providers/caldav/destination/a-create-conflict-never-overwrites-a-stranger.test.ts
```diff
@@ -0,0 +1,146 @@
...
+import { CalDAVCreateConflictError } from "../../../../src/providers/caldav/shared/client";
+import { eventToICalString } from "../../../../src/providers/caldav/shared/ics";
+
+/*
+ * Our object names are deterministic from the source event, so two keeper customers mirroring the
+ * same source event into the same shared calendar derive the same UID and therefore the same href.
+ * The second one's create conflicts, and the recovery then decided what to do about an object it
+ * did not put there. Deleting it and writing ours in its place destroys the first customer's event
+ * and reports conflictResolved. The same fall-through catches an object we merely could not parse.
+ */
```
ridafkih:
```suggestion
```

PR #980, opened 2026-08-31
fix(database): migrate timestamps on a server that was never UTC

### LINKED ISSUE #979: Upgrade error due to postgres timezone
I'm running Keeper via podman on Fedora, after upgrading from to `keeper-standalone:2.12.1` to `keeper-standalone:2.32.2`, Keeper shortly fails after starting due to:

```
401 |   const state = await connection.query<{ TimeZone: string }>("SHOW TimeZone");
402 |   const [row] = state.rows; 
403 |   if (isUtcTimeZoneName(row?.TimeZone)) {
404 |     return;
405 |   }
406 |   throw new Error(describeNonUtcTimeZone(row?.TimeZone));
                  ^
error: This database's server time zone is Europe/Berlin, not UTC. Timestamps written by column defaults carry that zone's wall clock, so migrating them to timestamptz would move every one of them by its offset. Set the Postgres server to UTC (ALTER SYSTEM SET timezone = 'UTC', then reload) and run this again.
      at <anonymous> (/app/packages/database/scripts/migrate.ts:406:13)

Bun v1.4.0 (Linux x64)
s6-rc: warning: unable to start service init-db: command exited 1 
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
/run/s6/basedir/scripts/rc.init: fatal: stopping the container.
s6-rc: info: service postgres: stopping
s6-rc: info: service redis: stopping

Session terminated, killing shell...2026-08-31 11:26:11.371 CEST [70] LOG:  received smart shutdown request
```

Probably introduced here https://github.com/ridafkih/keeper.sh/pull/821

ridafkih:
Fixes #979. The migration runner refused to start against a non-UTC Postgres, which bricked the standalone container on any host that wasn't UTC — including fresh installs with no data and databases that had already finished the conversion. It now rewrites the values it was worried about instead of refusing, and repairs the root cause so new clusters are UTC.

- `normalizeServerClockTimestamps` reads the server zone before the session is forced to UTC and moves the `DEFAULT now()` values to UTC while the columns are still naked `timestamp`; the bare ALTERs in 0087/0088 are then correct for both value origins. UTC servers return early and keep the existing no-rewrite fast path.
- Column origin comes off the schema, not the column name: a `$onUpdate` column is app-written, so only rows still equal to their table's insert-only column get moved. The three tables whose `updatedAt` has no sibling to date rows against are left alone rather than guessed at.
- `initdb --timezone=UTC` in the standalone image, so a new cluster stops inheriting the host zone.
- `server-time-zone-upgrade.test.ts` covers the four cases against real Postgres: not-yet-migrated and already-migrated, each on a UTC and a non-UTC server, plus an empty install. Verified red-first — the three non-UTC cases fail on `main` with the reported error.

Non-UTC users who already followed the old error message's advice have `createdAt`/`updatedAt` an offset in the future, and this doesn't find them — their columns are already `timestamptz`. Catching that group needs a separate one-shot migration and something we don't currently store.

https://claude.ai/code/session_017b1bFriu7qh2FAdEZ3VVZH


docker/standalone/rootfs/etc/s6-overlay/s6-rc.d/postgres/run
```diff
@@ -6,8 +6,10 @@ PGDATA=/var/lib/postgresql/data
 chown postgres:postgres "$PGDATA"
 chmod 0700 "$PGDATA"
 
+# initdb otherwise stamps the cluster with whatever zone the host handed the container,
+# and that zone is what DEFAULT now() writes into every timestamp column from then on.
```
ridafkih:
```suggestion
```

packages/database/scripts/migrate.ts
```diff
@@ -397,20 +401,49 @@ const isPostMigrationRuntimeReady = async (): Promise<boolean> => {
...
 };
 
-const assertUtcServerTimeZone = async (): Promise<void> => {
+/*
+ * Timestamps still typed `timestamp` hold a mix of UTC (written by the application) and
+ * server wall clock (written by DEFAULT now()). The conversion to timestamptz reads them
+ * all as UTC, so on a server that was never UTC the defaulted values have to be moved to
+ * UTC first. On a UTC server there is nothing to move, and on a database whose columns
+ * are already timestamptz there is nothing left to find.
+ */
```
ridafkih:
```suggestion
```

packages/database/src/database/server-clock-timestamps.ts
```diff
@@ -0,0 +1,121 @@
...
+  return siblings.find((column) => column === "createdAt") ?? siblings[0] ?? null;
+};
+
+const buildRewrite = (
+  { table, column }: ServerClockCandidate,
+  zone: string,
+): string =>
+  `UPDATE "${table}"`
+  + ` SET "${column}" = ("${column}" AT TIME ZONE '${zone}') AT TIME ZONE 'UTC'`
+  + ` WHERE "${column}" IS NOT NULL`;
```
ridafkih: fucking disgusting

packages/database/src/database/server-clock-timestamps.ts
```diff
@@ -0,0 +1,121 @@
...
+      const discriminator = findDiscriminator(candidate, candidates, origins);
+      if (!discriminator) {
+        return [];
+      }
+      return [
+        `${buildRewrite(candidate, zone)} AND "${candidate.column}" = "${discriminator}"`,
+      ];
+    }),
+    ...serverOnly.map((candidate) => buildRewrite(candidate, zone)),
+  ];
```
ridafkih: Good lord.

